PR #10, opened 2021-06-07
Fix unit in pd.to_timedelta()

frankenjoe:
Use `unit='s'`  instead of `'sec'` when converting seconds with `pd.to_timedelta()`.

Reason: it seems that some (older?) versions of `pandas` do not support `'sec'`, see e.g. https://github.com/audeering/opensmile-python/issues/32#issuecomment-850786628

### DIFF
--- a/audinterface/core/process.py
+++ b/audinterface/core/process.py
@@ -317,9 +317,7 @@ class Process:
         if start is None or pd.isna(start):
             start = pd.to_timedelta(0)
         if end is None or (pd.isna(end) and not self.keep_nat):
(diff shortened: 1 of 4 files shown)


hagenw:
Must indeed be older versions of `pandas`, compare:
* https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_timedelta.html
* https://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.to_timedelta.html

But I still think it is a good idea to change it.

Issue #14, opened 2021-06-29
process_signal_from_index() returns wrong index type for keep_nat


hagenw:
Try the following:

```python
import numpy as np
import pandas as pd

import audinterface


index = pd.MultiIndex.from_arrays([[pd.to_timedelta(0)], [pd.to_timedelta(pd.NaT)]], names=('start', 'end'))
process = audinterface.Process(keep_nat=True)
y = process.process_signal_from_index(np.zeros((1, 4)), 1, index)
y.index.get_level_values(1)
```

This returns
```
DatetimeIndex(['NaT'], dtype='datetime64[ns]', name='end', freq=None)
```

It's working if we don't request `keep_nat`:

```python
process = audinterface.Process()
y = process.process_signal_from_index(np.zeros((1, 4)), 1, index)
y.index.get_level_values(1)
```
which returns
```
TimedeltaIndex(['0 days 00:00:04'], dtype='timedelta64[ns]', name='end', freq=None)
```

frankenjoe:
I think the problem is that the `index` we input to the function is already has type `datetime` in the `end` level:

```python
index = pd.MultiIndex.from_arrays(
    [
        [pd.to_timedelta(0)],
        [pd.to_timedelta(pd.NaT)]
    ],
    names=('start', 'end'),
)
index.levels[1].dtype
```
```
datetime64[ns]
```

frankenjoe:
It works when you call `pd.to_timedelta()` on the whole array, i.e.:

```python
index = pd.MultiIndex.from_arrays(
    [
        pd.to_timedelta([0]),
        pd.to_timedelta([pd.NaT])
    ],
    names=('start', 'end'),
)
process = audinterface.Process(keep_nat=True)
y = process.process_signal_from_index(np.zeros((1, 4)), 1, index)
y.index.get_level_values(1)
```
```
DatetimeIndex(['NaT'], dtype='datetime64[ns]', name='end', freq=None)
```

So I would say it's not a bug, because the user is supposed to pass values of type `datetime` (though I agree that `pandas` is given us a hard time here).

hagenw:
`pandas` makes it so hard that at some point I would call it a bug, and I regret every day that we are using it ;)

Is there a way for us to convert the incoming index values to `TimeDelta` ourselves?
Before I tried
```python
index.get_level_values(1).map(lambda x: pd.to_timedelta(x))
```
but it did not work.

hagenw:
[...]

PR #17, opened 2021-07-05
Use keyword argument with MultiIndex.set_levels()

### LINKED ISSUE #16: Fix pandas future warnings
With the current release of `audinterface` we get warnings like:

```
/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/audinterface/core/process.py:127: FutureWarning: In a future version of pandas all arguments of MultiIndex.set_levels except for the argument 'levels' will be keyword-only
30
  y.index = y.index.set_levels([[start], [end]], [1, 2])
```

As this can lead to [failing tests](https://github.com/audeering/opensmile-python/pull/39/checks?check_run_id=2986991383), we should update the code.

frankenjoe:
Closes #16

### DIFF
--- a/audinterface/core/process.py
+++ b/audinterface/core/process.py
@@ -124,7 +124,10 @@ class Process:
         if end is None or (pd.isna(end) and not self.keep_nat):
             end = y.index.levels[2][0] + start
 
-        y.index = y.index.set_levels([[start], [end]], [1, 2])
+        y.index = y.index.set_levels(
+            [[start], [end]],
+            level=[1, 2],
+        )
 
         return y
 
--- a/audinterface/core/segment.py
+++ b/audinterface/core/segment.py
@@ -378,7 +378,9 @@ class Segment:
                 [
                     index.levels[0] + start,
                     index.levels[1] + start,
-                ], [0, 1])
+                ],
+                level=[0, 1],
+            )
         if file is not None:
             index = pd.MultiIndex(
                 levels=[


frankenjoe:
We now have a another problem:

```
  /home/runner/work/audinterface/audinterface/audinterface/core/feature.py:125: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
```

I don't get the error with `Python 3.6` and latest `numpy`...

hagenw:
We even get an error for our famous time objects:

```
       AssertionError: MultiIndex level [1] are different
341
E       
342
E       MultiIndex level [1] classes are not equivalent
343
E       [left]:  DatetimeIndex(['NaT'], dtype='datetime64[ns]', name='end', freq=None)
344
E       [right]: Float64Index([nan], dtype='float64', name='end')
```

frankenjoe:
> We even get an error for our famous time objects:

yes, seems actually be related to https://github.com/audeering/audinterface/issues/14, so maybe I'll prepare a PR for that issue first.

frankenjoe:
@hagenw cool, after rebasing the error indeed disappeared!

hagenw:
Great, so this now works with Python 3.6 and higher versions.

PR #18, opened 2021-07-06
Add utils.signal_index()

### LINKED ISSUE #14: process_signal_from_index() returns wrong index type for keep_nat

frankenjoe:
Closes #14 

Introduces utility function `audinterface.utils.signal_index()`, which creates a segmented index just like  `audformat.segmented_index()` but without the `file` level. It now takes care of using the correct `Timedelta` dtype.

The original example in #14 now fails:

```python
index = pd.MultiIndex.from_arrays(
    [
        [pd.to_timedelta(0)],
        [pd.to_timedelta(pd.NaT)]
    ],
    names=('start', 'end'),
)
process = audinterface.Process(keep_nat=True)
y = process.process_signal_from_index(np.zeros((1, 4)), 1, index)
y.index.get_level_values(1)
```
```
ValueError: Level 'end' must contain values of type 'timedelta64[ns]'.
```

and the fixed example has the correct type:

```python
index = pd.MultiIndex.from_arrays(
    [
        pd.to_timedelta([0]),
        pd.to_timedelta([pd.NaT])
    ],
    names=('start', 'end'),
)
process = audinterface.Process(keep_nat=True)
y = process.process_signal_from_index(np.zeros((1, 4)), 1, index)
y.index.get_level_values(1)
```
```
TimedeltaIndex([NaT], dtype='timedelta64[ns]', name='end', freq=None)
```

and can be simplified to:

```python
index = audinterface.utils.signal_index(0, pd.NaT)
process = audinterface.Process(keep_nat=True)
y = process.process_signal_from_index(np.zeros((1, 4)), 1, index)
y.index.get_level_values(1)
```
```
TimedeltaIndex([NaT], dtype='timedelta64[ns]', name='end', freq=None)
```

### DIFF
--- a/audinterface/core/utils.py
+++ b/audinterface/core/utils.py
@@ -163,6 +163,12 @@ def signal_index(
     Returns a segmented index like
     :func:`audformat.segmented_index`,
     but without the ``'file'`` level.
+    Can be used with the following methods:
+
+    * :meth:`audinterface.Feature.process_signal_from_index`
+    * :meth:`audinterface.Process.process_signal_from_index`
+    * :meth:`audinterface.ProcessWithContext.process_signal_from_index`
+    * :meth:`audinterface.Segment.process_signal_from_index`
 
     Args:
         starts: segment start positions.
--- a/audinterface/core/utils.py
+++ b/audinterface/core/utils.py
@@ -160,6 +160,10 @@ def signal_index(
 ) -> pd.MultiIndex:
     r"""Create signal index.
 
+    Returns a segmented index like
+    :func:`audformat.segmented_index`,
+    but without the ``'file'`` level.
+
     Args:
         starts: segment start positions.
             Time values given as float or integers are treated as seconds
@@ -167,10 +171,7 @@ def signal_index(
             Time values given as float or integers are treated as seconds
 
     Returns:
-        signal index
-
-    Raises:
-        ValueError: if created index contains duplicates
+        index with start and end times
 
     Raises:
         ValueError: if ``start`` and ``ends`` differ in size
@@ -192,6 +193,14 @@ def signal_index(
         ... )
         MultiIndex([('0 days 00:00:00', '0 days 00:00:01'),
                     ('0 days 00:00:01', '0 days 00:00:02')],
+                   names=['start', 'end'])
+        >>> signal_index([0, 1])
+        MultiIndex([('0 days 00:00:00', NaT),
+                    ('0 days 00:00:01', NaT)],
+                   names=['start', 'end'])
+        >>> signal_index(ends=[1, 2])
+        MultiIndex([('0 days', '0 days 00:00:01'),
+                    ('0 days', '0 days 00:00:02')],
                    names=['start', 'end'])
 
     """
--- a/audinterface/core/utils.py
+++ b/audinterface/core/utils.py
@@ -12,24 +12,40 @@ import audiofile as af
 from audinterface.core.typing import Timestamps
 
 
-def check_index(index: pd.Index):
+def assert_index(obj: pd.Index):
     r"""Check if index is conform to audformat."""
-    if isinstance(index, pd.MultiIndex) and len(index.levels) == 2:
-        if not index.empty:
-            if not pd.core.dtypes.common.is_datetime_or_timedelta_dtype(
-                    index.levels[0]
-            ):
-                raise ValueError(f'Level 0 has type '
-                                 f'{type(index.levels[0].dtype)}'
-                                 f', expected timedelta64[ns].')
-            if not pd.core.dtypes.common.is_datetime_or_timedelta_dtype(
-                    index.levels[1]
-            ):
-                raise ValueError(f'Level 1 has type '
-                                 f'{type(index.levels[1].dtype)}'
-                                 f', expected timedelta64[ns].')
+
+    if isinstance(obj, pd.MultiIndex) and len(obj.levels) == 2:
+        if not (
+                obj.names[0] == audformat.define.IndexField.START
+                and obj.names[1] == audformat.define.IndexField.END
+        ):
+            expected_names = [
+                audformat.define.IndexField.START,
+                audformat.define.IndexField.END,
+            ]
+            raise ValueError(
+                'Found two levels with names '
+                f'{obj.names}, '
+                f'but expected names '
+                f'{expected_names}.'
+            )
+        if not pd.api.types.is_timedelta64_dtype(obj.levels[0].dtype):
+            raise ValueError(
+                "Level 'start' must contain values of type 'timedelta64[ns]'."
+            )
+        if not pd.api.types.is_timedelta64_dtype(obj.levels[1].dtype):
+            raise ValueError(
+                "Level 'end' must contain values of type 'timedelta64[ns]'."
+            )
     else:
-        audformat.assert_index(index)
+        audformat.assert_index(obj)
+
+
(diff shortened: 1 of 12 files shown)


audinterface/core/utils.py
```diff
@@ -138,6 +154,90 @@ def segments_to_indices(
...
+            Time values given as float or integers are treated as seconds
+
+    Returns:
+        signal index
+
+    Raises:
+        ValueError: if created index contains duplicates
+
+    Raises:
+        ValueError: if ``start`` and ``ends`` differ in size
```
hagenw, line 176:
```suggestion
    Raises:
        ValueError: if created index contains duplicates
        ValueError: if ``start`` and ``ends`` differ in size
```

audinterface/core/utils.py
```diff
@@ -138,6 +154,90 @@ def segments_to_indices(
...
+
+    index = pd.MultiIndex.from_arrays(
+        [
+            to_timedelta(starts),
+            to_timedelta(ends),
+        ],
+        names=[
+            audformat.define.IndexField.START,
+            audformat.define.IndexField.END,
+        ])
```
hagenw, line 225:
```suggestion
        ]
    )
```

audinterface/core/utils.py
```diff
@@ -138,6 +154,90 @@ def segments_to_indices(
...
+    assert_index(index)
+
+    return index
+
+
+def to_array(value: typing.Any) -> typing.Union[list, np.ndarray]:
+    r"""Convert value to list or array."""
```
hagenw, line 232: Why not only to array?
E.g.
```python
elif is_scalar(value):
    value = np.array([value])
```

hagenw:
Great, every time we don't have to use `pd.MultiIndex()` ourselves makes things much easier to understand.

I'm wondering if we should name it `time_index` instead of `signal_index`?

audinterface/core/utils.py
frankenjoe: Actually, we don't check for duplicates, so I removed it. Since we do not deal with annotations here, I don't we have to check for duplicates.

frankenjoe:
> I'm wondering if we should name it `time_index` instead of `signal_index`?

I thought `signal_index` connects better to `process_signal_from_index()`.

audinterface/core/utils.py
frankenjoe: changed it

hagenw:
> > I'm wondering if we should name it `time_index` instead of `signal_index`?
> 
> I thought `signal_index` connects better to `process_signal_from_index()`.

I had `time_index()` in mind, because I first thought we are inside `audformat` here, where we define `segmented_index()`.
I thing both would work with `process_signal_from_index()`

audinterface/core/utils.py
frankenjoe: makes sense, changed it.

frankenjoe:
> I had `time_index()` in mind, because I first thought we are inside `audformat` here, where we define `segmented_index()`

Mhh, but actually it's also a `segmented_index()`, with the difference that it does not have a `file` level. I.e. you cannot apply it to files but only to signals.

audinterface/core/utils.py
hagenw: You can get `pandas` errors for duplicated index entries, but it depends what you are doing. The advantage of duplicates is that we can force an object to process the same segment more than one time.

hagenw:
> > I had `time_index()` in mind, because I first thought we are inside `audformat` here, where we define `segmented_index()`
> 
> Mhh, but actually it's also a `segmented_index()`, with the difference that it does not have a `file` level. I.e. you cannot apply it to files but only to signals.

OK, then let's stay with `signal_index()`.

audinterface/core/utils.py
hagenw: But there is one downside, further down you set `signal_index` is also a `segmented_index` only without the file level. This would then be no longer true, as a segmented index is not allowed to contain duplicates.

frankenjoe: So should we raise an error for duplicates then?

frankenjoe:
> OK, then let's stay with `signal_index()`.

In the docstrings of `signal_index()` I added links to the `*.process_signal_from_index()` functions and vice versa.

audinterface/core/utils.py
hagenw: I'm still not completely sure.
Maybe it is handy to allow for duplicated index entries.

The advantage of not allowing duplicates is that you only have to add the file level and you have a segmented index.

Any thoughts what would be more important?

frankenjoe: Maybe it's better to forbid duplicates and later allow them if we see the necessity, as the other way round would be a breaking change?

hagenw: Good argument, then let's first forbid duplicates.

frankenjoe: Mhh just realized that we have the following code in `Feature`:

```python
        if n_time_steps > 1:
            if win_dur is None:
                starts = [start] * n_time_steps
                ends = [end] * n_time_steps
```

so basically if we have a process function that returns more than one frame, but `win_dur` is set to `None` we duplicate `start` and `end` for each frame. This would not be possible any longer if we do not allow for duplicates. On the other hand, this will also create a bad `audformat`  index when applied on files I think....

hagenw: Yes, sounds we shouldn't doing this.
I would say hat we should not allow for `win_dur` `None` in this case, or we have to return only one time step and ignore the `n_time_steps` setting.

frankenjoe: Yes, we indeed get an error as the following minimal example shows:

```python
import audinterface


    ['f1', 'f2'],
    process_func=lambda x, sr: np.ones((1, 2, 3))  # return 3 frames
)

index = audformat.segmented_index('test.wav')

feature.process_index(index)
```
```
ValueError: Index not conform to audformat. Found duplicates:
('test.wav', Timedelta('0 days 00:00:00'), Timedelta('0 days 00:00:00.000187500'))
('test.wav', Timedelta('0 days 00:00:00'), Timedelta('0 days 00:00:00.000187500'))
```

With `process_signal_from_index()` this still works and that's why do not get an error in our tests at the moment.

frankenjoe: > I would say hat we should not allow for win_dur None in this case, or we have to return only one time step and ignore the n_time_steps setting.

Ok, this will be a breaking change, thought I doubt we have code that will break so far. So should I continue in this PR or do it in a separate one?

hagenw: It's a breaking change, but the current behavior is simply a bug.

I will merge the current stage and propose to fix the bug in another pull request and also forbid duplicates for `signal_index()` in another pull request.

Issue #21, opened 2021-07-07
Feature from (list of) floats and arrays with one dimension


frankenjoe:
The processing function of a `Feature` object currently has to return a numpy array with 2 or 3 dimensions. I wonder if we should lose this constraint and also support:

1. single float value or list of float values
2. numpy arrays with a single dimension

currently, if we have two dimension we already assume that the third dimension (i.e. the number of frames) is 1. In addition we could assume that for a plain list of values or an array with a single dimension the number of channels is 1.

I think this makes especially sense as we already have a flag `process_func_is_mono`, where we apply the processing function to all available channels, but at the moment we still require that the process function returns the number of channels as first dimension.

It will also make it a bit easier for a user to write a process function, e.g. to calculate the mean and max of signal, she can then write:

```python
def process_func(x, sr):
    return [x.min(), x.max()]
```

instead of:

```python
def process_func(x, sr):
    return np.atleast2d([x.min(), x.max()])
```

or to calculate the mean, simply:

```python
def process_func(x, sr):
    return x.mean()
```

hagenw:
Do you remember what was the reason why we introduced this requirement for `Feature` as for `Process` we don't have it.

frankenjoe:
`Process` returns a `pd.Series` and can hold any type of elements, e.g. strings, tuples, anything. But with `Feature` we create a `pd.DataFrame` and the return type has to be something we can expand into the columns, so we have to make sure we always get the same shape, which we defined as `(channels, features, frames)`. But we already allow also `(channels, features)`, which we then expand to `(channels, features, 1)`  and here I suggest to also allow `(features)`, which we then expand to `(1, features, 1)`.

hagenw:
OK, I think then one reason to require channel is to avoid the problem of deciding what to do when you have a process functino that returns only one channel, but then starts doing windowing, because then we get `(features, frames)` which we cannot differentiate from `(channels, features)`

frankenjoe:
Well, the rule would be - if we get a 1d array the values hold only the features and we set channels to 1 and frames to 1. If we get a 2d array the values hold channels and features and we set frames to 1. The latter we already do now.

PR #22, opened 2021-07-14
Support more feature shapes

frankenjoe:
As discussed in #21 it would be nice to be less strict on the returned shape of the process function that can be used with `Feature`. Especially since we already support 2d, but in that case the first dimension always has to be the channels, which does not always make sense, e.g. with `process_func_is_mono=True` we already know that the channel dimension will be 1.

Luckily, since `channels` and `features` are fixed and only `frames` can vary, we can derive the correct 3d shape from 1 or 2 dimensions. In this PR we add support for:

* `(features)`
* `(features, frames)`

in addition to:

* `(channels, features)`
* `(channels, features, frames)`

## Bug fix

While implementing this PR I realized that `__call__()` did not return `(channels, features, frames)` as promised, but the original shape from the processing function (which does not make sense, cause if that is the intension of the user she should use `Process`). This is now fixed (we probably didn't find the bug yet, because we seldomly use `__call__()` but rather one of the process functions, which were not affected).

### Single frame on mono file

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random(2),
)
```

or

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((1, 2)),
)
```

or

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((2, 1)),
)
```

or

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((1, 2, 1)),
)
```

return:

```python
feature.process_file('mono.wav')
```
```
                                                  x         y
file     start  end                                          
mono.wav 0 days 0 days 00:00:07.247709751  0.017839  0.185667
```

### Single frame on stereo file

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((2, 2)),
    channels=[0, 1],
)
```

or

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((2, 2, 1)),
    channels=[0, 1],
)
```

or

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random(2),
    channels=[0, 1],
    process_func_is_mono=True,
)
```

or

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((1, 2)),
    channels=[0, 1],
    process_func_is_mono=True,
)
```

or

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((2, 1)),
    channels=[0, 1],
    process_func_is_mono=True,
)
```

or

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((1, 2, 1)),
    channels=[0, 1],
    process_func_is_mono=True,
)
```

return:

```python
feature.process_file('stereo.wav')
```
```
                                                  x-0       y-0       x-1       y-1
file       start  end                                                              
stereo.wav 0 days 0 days 00:00:07.247709751  0.353643  0.968675  0.304591  0.965526
```

### Multiple frames on mono file

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((2, 3)),
    win_dur=1.0,
)
```

or 

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((1, 2, 3)),
    win_dur=1.0,
)
```

return

```python
feature.process_file('mono.wav')
```
```
                                                               x         y
file     start                  end                                       
mono.wav 0 days 00:00:00        0 days 00:00:01         0.594710  0.465447
         0 days 00:00:00.500000 0 days 00:00:01.500000  0.800685  0.865276
         0 days 00:00:01        0 days 00:00:02         0.574075  0.012008
```

### Multiple frames on stereo file

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((2, 2, 3)),
    channels=[0, 1],
    win_dur=1.0,
)
```

or 

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((2, 3)),
    channels=[0, 1],
    process_func_is_mono=True,
    win_dur=1.0,
)
```

or 

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((1, 2, 3)),
    channels=[0, 1],
    process_func_is_mono=True,
    win_dur=1.0,
)
```

return:

```python
feature.process_file('stereo.wav')
```
```
                                                               x-0       y-0       x-1       y-1
file       start                  end                                                           
stereo.wav 0 days 00:00:00        0 days 00:00:01         0.216335  0.201805  0.907083  0.913285
           0 days 00:00:00.500000 0 days 00:00:01.500000  0.078636  0.977286  0.682514  0.933484
           0 days 00:00:01        0 days 00:00:02         0.245939  0.684865  0.169649  0.852454
```

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -457,7 +457,7 @@ class Feature:
                 f'not {features.ndim}.'
             )
 
-        # reshape features to (channels x features x frames)
+        # figure out channels, feature, frames
         if features.ndim == 1:
             n_channels = 1
             n_features = features.size
@@ -484,8 +484,7 @@ class Feature:
             n_features = features.shape[1]
             n_frames = features.shape[2]
 
-        features = features.reshape([n_channels, n_features, n_frames])
-
+        # assert channels and features have expected length
         if n_channels != self.num_channels:
             raise RuntimeError(
                 f'Number of channels must be'
@@ -501,7 +500,8 @@ class Feature:
                 f'{n_features}.'
             )
 
-        return features
+        # reshape features to (channels,  features, frames)
+        return features.reshape([n_channels, n_features, n_frames])
 
     def _series_to_frame(
             self,
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -560,7 +560,8 @@ class Feature:
         features = self._reshape_3d(features)
         n_channels, n_features, n_frames = features.shape
 
-        # Reshape features and store channel number as first feature
+        # Combine features and channels into one dimension
+        # f1-c1, f2-c1, fN-c1, ..., f1-cM, f2-cM,..., fN-cM
         new_shape = (n_channels * n_features, n_frames)
         features = features.reshape(new_shape).T
 
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -415,6 +415,77 @@ class Feature:
         """
         return frame.values.T.reshape(self.num_channels, self.num_features, -1)
 
+    def _assert_shape_3d(
+            self,
+            features: typing.Union[np.ndarray, pd.Series]
+    ):
+        r"""Reshape to [n_channels, n_features, n_frames]."""
+
+        if self.process.process_func_is_mono:
+            features = np.array(features)
+            if features.ndim == 4 and features.shape[1] == 1:
+                # mono processing turned on
+                # and process function returns (1, features, frames)
+                features = features.squeeze(axis=1)
+
+        if not isinstance(features, np.ndarray):
+            raise RuntimeError(
+                "Features must be a 'np.ndarray', "
+                f"not '{type(features)}'."
+            )
+
+        if features.ndim > 3:
+            raise RuntimeError(
+                f'Dimension of extracted features must be 1, 2 or 3, '
+                f'not {features.ndim}.'
+            )
+
+        # reshape features to (channels x features x frames)
+        if features.ndim == 1:
+            n_channels = 1
+            n_features = features.size
+            n_frames = 1
+        elif features.ndim == 2:
+            if (features.shape[0] == self.num_channels) and \
+                    (features.shape[1] == self.num_features):
+                n_channels = features.shape[0]
+                n_features = features.shape[1]
+                n_frames = 1
+            elif features.shape[0] == self.num_features:
+                n_channels = 1
+                n_features = features.shape[0]
+                n_frames = features.shape[1]
+            else:
+                raise RuntimeError(
+                    f'Cannot determine feature shape from '
+                    f'{features.shape}, ',
+                    f'when expected shape is '
+                    f'({self.num_channels, self.num_features, -1}).'
+                )
+        else:
+            n_channels = features.shape[0]
+            n_features = features.shape[1]
+            n_frames = features.shape[2]
+
+        features = features.reshape([n_channels, n_features, n_frames])
+
+        if n_channels != self.num_channels:
+            raise RuntimeError(
+                f'Number of channels must be'
+                f' {self.num_channels}, '
+                f'not '
+                f'{n_channels}.'
+            )
+        if n_features != self.num_features:
+            raise RuntimeError(
+                f'Number of features must be '
+                f'{self.num_features}, '
+                f'not '
+                f'{n_features}.'
+            )
+
+        return features
+
     def _series_to_frame(
             self,
             series: pd.Series,
@@ -448,10 +519,11 @@ class Feature:
     ) -> pd.DataFrame:
 
         # Convert features to a pd.DataFrame
-        # Assumed format
-        # [n_channels, n_features, n_time_steps]
-        # or
+        # Assumed formats are:
+        # [n_channels, n_features, n_frames]
         # [n_channels, n_features]
+        # [n_features, n_frames]
+        # [n_features]
 
         if self.win_dur is not None:
             if self.unit == 'samples':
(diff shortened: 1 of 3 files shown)


hagenw:
I think the following behavior is wrong:

```python
    ['x', 'y'],
    process_func=lambda x, sr: np.random.random((2, 1)),
)
```

returns

```
                                                  x         y
file     start  end                                          
mono.wav 0 days 0 days 00:00:07.247709751  0.017839  0.185667
```

as our default in `Feature` is `process_func_is_mono=False`. So how can I know that the first entry of the returned value of the process function is not the channel number? E.g. the process function could do an upmix from mono to stereo and return `(channel, feature)` instead of `(feature, frame)` that you silently assume here.

frankenjoe:
> So how can I know that the first entry of the returned value of the process function is not the channel number?

Because we know that we have 2 features, so only the first dimension can be the features.

hagenw:
> > So how can I know that the first entry of the returned value of the process function is not the channel number?
> 
> Because we know that we have 2 features, so the first dimension has to be the features.

Ah, ok, that makes sense.Because a user has to specify the feature names, we know how many features there are and because of `channels` we know how many channels there are, correct?

frankenjoe:
> Ah, ok, that makes sense.Because a user has to specify the feature names, we know how many features there are and because of `channels` we know how many channels there are, correct?

Exactly!

You may want to have a look at the `_assert_shape_3d()` function, which contains the logic that tries to figure out the correct mapping if we have only 1 or 2 dimensions.

audinterface/core/feature.py
```diff
@@ -415,6 +417,92 @@ def to_numpy(
...
+            # the channel dimension has to be 1
+            # so we would usually omit it,
+            # but since older versions required
+            # a channel dimension we have to
+            # consider two special cases
+            if (features.ndim == 4) and \
+                    (features.shape[1] == 1):
+                # (1, features, frames)
+                # -> (channels, 1, features, frames)
+                # -> (channels, features, frames)
```
hagenw, line 438: This comment is slightly misleading.
What you are trying to do here is to fix
```
(channels, channels, features, frames)
```
to
```
(channels, features, frames)
```
correct?

audinterface/core/feature.py
```diff
@@ -415,6 +417,92 @@ def to_numpy(
...
+                features = features.squeeze(axis=1)
+            elif (features.ndim == 3) and \
+                    (self.win_dur is None) and \
+                    (features.shape[1] == 1):
+                # (1, features)
+                # -> (channels, 1, features)
+                # -> (channels, features)
```
hagenw, line 445: And here
```
(channels, channels, features)
```
to
```
(channels, features)
```

So maybe change the comment to
```python
# (channels, channels, features)
# -> (channels, features)
```

audinterface/core/feature.py
hagenw: I also don't complete get, why we can have `(channels, channels, features, frames)` in the first place?

audinterface/core/feature.py
```diff
@@ -415,6 +417,92 @@ def to_numpy(
...
         return frame.values.T.reshape(self.num_channels, self.num_features, -1)
 
+    def _assert_shape_3d(
+            self,
+            features: typing.Union[np.ndarray, pd.Series]
+    ):
+        r"""Reshape to [n_channels, n_features, n_frames]."""
+
+        if self.process.process_func_is_mono:
+            features = np.array(features)
```
hagenw, line 427: If we don't have `self.process.process_func_is_mono` we raise an error if `features` is not a numpy array.
Why is it ok here if we don't have one?

audinterface/core/feature.py
```diff
@@ -468,58 +557,27 @@ def _values_to_frame(
...
-            raise RuntimeError(
-                f'Number of channels must be {self.num_channels}, '
-                f'not {n_channels}.'
-            )
-        if n_features != len(self.feature_names):
-            raise RuntimeError(
-                f'Number of features must be {len(self.feature_names)}, '
-                f'not {n_features}.'
-            )
+        features = self._assert_shape_3d(features)
```
hagenw, line 560: Maybe not call it `_assert_shape_3d` as it will not return `True` or `False` but will change the features.

audinterface/core/feature.py
```diff
@@ -468,58 +557,27 @@ def _values_to_frame(
...
+        n_channels, n_features, n_frames = features.shape
 
         # Reshape features and store channel number as first feature
         # [n_channels, n_features, n_time_steps] =>
         # [n_channels * n_features + 1, n_time_steps]
```
hagenw, line 565: Looking at the code the result should be
```
# [n_channels * n_features, n_time_steps]
```
and maybe also replace `n_time_steps` by `n_frames` as in the code?

hagenw: As we can also clearly read from the code line how the new shape looks like it might also be sufficient if the comment just reads:
```python
# Reshape features and store channel number as first feature
```

audinterface/core/feature.py
frankenjoe: It's for the special case that `process_func_is_mono=True`. In that case we get a list of `numpy.arrays` we need to combine to an array of shape `(channels, features, frames)`. Now, when a user sets `process_func_is_mono=True` it makes sense to return `(features, frames)` since the process function only gets mono audio. But of course we don't want to forbid that it returns `(1, features, frames)` (actually, before this MR it even had to return this shape). But that means that when we combine the arrays we get `(channels, 1, features, frames)`, so in that case we have to remove the second dimension.

audinterface/core/feature.py
frankenjoe: Like before but for the case that the we do not process frames.

audinterface/core/feature.py
frankenjoe: The process function still has to return `np.array`, but internally we loop over all channels and collect the results in a list and here we convert that list to a `np.array`, i.e. we combine the results of the channels to a single array.

audinterface/core/feature.py
hagenw: Thanks for the detailed explanation.

Then maybe change the comment to:

```python
# (1, features, frames)  mono feature extractor
# -> (channels, 1, features, frames)  combined extracted mono features
# -> (channels, features, frames)  desired shape
```

or

```python
# (channels, 1, features, frames)
# -> (channels, features, frames)
```

audinterface/core/feature.py
hagenw: But will it still raise an error in this case if the process function does not return `np.array` and we have `process_func_is_mono`?

audinterface/core/feature.py
frankenjoe: Actually the comment is also misleading. I think in a very first draft of `Features` we had the idea to encode the channel number as a feature, but then we changed it and instead we encode it in the column name of the `pd.DataFrame`. I now changed the comment to:

```
# Combine features and channels into one dimension
# f1-c1, f2-c1, ..., fN-c1, ..., f1-cM, f2-cM, ..., fN-cM
```

audinterface/core/feature.py
frankenjoe: Renamed to `_reshape_3d()`.

audinterface/core/feature.py
frankenjoe: Ok, changed it accordingly.

audinterface/core/feature.py
frankenjoe: I changed the comment to:

```
# (channels, 1, features)
# -> (channels, features)
```

audinterface/core/feature.py
frankenjoe: Oh now I get your point. Yes, I think you are right, we actually need to assert that every element of the list is a `np.ndarray`.

audinterface/core/feature.py
hagenw: Yes, much better, thanks.

audinterface/core/feature.py
frankenjoe: I fixed it and added a test for it.

hagenw: Great, looks good now.

PR #23, opened 2021-07-22
Feature: support scalars and lists

### LINKED ISSUE #21: Feature from (list of) floats and arrays with one dimension

frankenjoe:
Closes #21

Add support for to scalars and lists to `Feature`.

### Examples

```python
    ['x'],
    process_func=lambda x, sr: 1,
)
feature.process_file('mono.wav')
```
```
                                           x
file     start  end                         
mono.wav 0 days 0 days 00:00:07.247709751  1
```
```python
    ['x', 'y'],
    process_func=lambda x, sr: [1, 2],
)
 feature.process_file('mono.wav')
```
```
                                           x  y
file     start  end                            
mono.wav 0 days 0 days 00:00:07.247709751  1  2
```
```pythonfeature = audinterface.Feature(
    ['x', 'y'],
    process_func=lambda x, sr: [[1, 1, 1], [2, 2, 2]],
    win_dur=1.0,
)
feature.process_file('mono.wav')
```
```
                                                        x  y
file     start                  end                         
mono.wav 0 days 00:00:00        0 days 00:00:01         1  2
         0 days 00:00:00.500000 0 days 00:00:01.500000  1  2
         0 days 00:00:01        0 days 00:00:02         1  2
```

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -94,7 +94,7 @@ class Feature:
             *,
             name: str = None,
             params: typing.Dict = None,
-            process_func: typing.Callable[..., np.ndarray] = None,
+            process_func: typing.Callable[..., typing.Any] = None,
             process_func_is_mono: bool = False,
             sampling_rate: int = None,
             win_dur: typing.Union[int, float] = None,
@@ -423,14 +423,10 @@ class Feature:
     ):
         r"""Reshape to [n_channels, n_features, n_frames]."""
 
+        features = np.array(features)
+        features = np.atleast_1d(features)
+
         if self.process.process_func_is_mono:
-            for channels_features in features:
-                if not isinstance(channels_features, np.ndarray):
-                    raise RuntimeError(
-                        "Features must be a 'np.ndarray', "
-                        f"not '{type(channels_features)}'."
-                    )
-            features = np.array(features)
             # when mono processing is turned on
             # the channel dimension has to be 1
             # so we would usually omit it,
@@ -448,12 +444,6 @@ class Feature:
                 # (channels, 1, features)
                 # -> (channels, features)
                 features = features.squeeze(axis=1)
-
-        if not isinstance(features, np.ndarray):
-            raise RuntimeError(
-                "Features must be a 'np.ndarray', "
-                f"not '{type(features)}'."
-            )
 
         if features.ndim > 3:
             raise RuntimeError(
--- a/tests/test_feature.py
+++ b/tests/test_feature.py
@@ -302,6 +302,24 @@ def test_process_folder(tmpdir):
         ),
         # 1 channel, 1 feature
         (
+            lambda s, sr: 1,
+            1,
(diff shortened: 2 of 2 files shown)


hagenw:
BTW, your changes do not only suppoort scalars:

```python
    ['x'],
    process_func=lambda x, sr: 'a',
)
feature.process_file('test.flac')
```
```
                                  x
file      start  end               
test.flac 0 days 0 days 00:00:15  a
```

```python
    ['x', 'y'],
    process_func=lambda x, sr: ['a', 'b'],
)
feature.process_file('test.flac')
```
```
                                  x  y
file      start  end                  
test.flac 0 days 0 days 00:00:15  a  b
```

which is nice, but I wanted to mention it.

frankenjoe:
Yes, I also noticed that it now works with any type, e.g.:

```python
    ['x', 'y'],
    process_func=lambda x, sr: [datetime.datetime.now()] * 2,
)
feature.process_file('mono.wav')
```
```
                                                                   x                          y
file     start  end                                                                            
mono.wav 0 days 0 days 00:00:07.247709751 2021-07-23 09:50:51.316974 2021-07-23 09:50:51.316974
```

I guess it's because `np.array()` works with any type.

audinterface/core/feature.py
```diff
@@ -423,14 +423,10 @@ def _reshape_3d(
...
 
+        features = np.array(features)
+        features = np.atleast_1d(features)
+
         if self.process.process_func_is_mono:
-            for channels_features in features:
-                if not isinstance(channels_features, np.ndarray):
-                    raise RuntimeError(
-                        "Features must be a 'np.ndarray', "
-                        f"not '{type(channels_features)}'."
```
hagenw, line 431: Seems like we haven't neither mentioned this error in docs not stated that we need to return a numpy array.
I was just wondering about that, but couldn't find an entry in the API docs (we don't have user documentation anyway ;) ).

hagenw:
Good, that should provide us with a lot of flexibility and we should be able to convert anything into a feature now, which I think is not bad.

PR #24, opened 2021-07-23
DOC: fix typo in Feature docstring

hagenw:
Changes

image

to

image


hagenw:
There was an error when I uploaded the first image in the description. It's now fixed and you can see the difference between the two images.

PR #33, opened 2022-01-04
Release 0.8.0

hagenw:
Before doing a 1.0.0 release we need to do a 0.8.0 release in between as we have marked stuff as deprecated and to be removed with 1.0.0 (https://github.com/audeering/audinterface/pull/28), and we have removed stuff that was marked to be removed in 0.8.0 (https://github.com/audeering/audinterface/pull/29).


hagenw:
I'm afraid this will also once again break some of our existing code, but we are now very near to version 1.0.0 ;)

PR #34, opened 2022-01-19
Process_folder(): handle empty and non-existing folder

### LINKED ISSUE #35: Default dtype of empty series / frame
When we return an empty series or frame we need to set a `dtype`. Currently we use `float` for in case of a series (`Process`) and `object` in case of a frame (`Feature`). The default type of `pandas` will be `object` in the future. Maybe we should simply follow their solution?

frankenjoe:
Closes #35 

If a folder is empty or does not exist, `process_folder()` currently raises a `ValueError: All objects passed were None`.

This changes the behavior of `process_folder()` to:

* Raise `FileNotFoundError` if folder does not exist. 
* Return an empty series/frame if folder is empty.

Other change: when an empty series or frame is returned, the data type is set to `object`.

### DIFF
--- a/tests/test_feature.py
+++ b/tests/test_feature.py
@@ -293,9 +293,9 @@ def test_process_folder(tmpdir):
 
     # empty folder
     root = str(tmpdir.mkdir('empty'))
-    y = feature.process_folder(root)
+    df = feature.process_folder(root)
     pd.testing.assert_frame_equal(
-        y,
+        df,
         pd.DataFrame(
             dtype=float,
             columns=feature.column_names,
--- a/tests/test_feature.py
+++ b/tests/test_feature.py
@@ -297,7 +297,7 @@ def test_process_folder(tmpdir):
     pd.testing.assert_frame_equal(
         y,
         pd.DataFrame(
-            dtype=object,
+            dtype=float,
             columns=feature.column_names,
         ),
     )
--- a/tests/test_feature.py
+++ b/tests/test_feature.py
@@ -262,8 +262,9 @@ def test_process_file(tmpdir, start, end, segment):
 def test_process_folder(tmpdir):
 
     index = audinterface.utils.signal_index(0, 1)
+    feature_names = ['o1', 'o2', 'o3']
     feature = audinterface.Feature(
-        feature_names=('o1', 'o2', 'o3'),
+        feature_names,
         process_func=feature_extractor,
         sampling_rate=None,
         channels=range(NUM_CHANNELS),
@@ -285,6 +286,21 @@ def test_process_folder(tmpdir):
     assert all(y.index.levels[1] == index.levels[0])
     assert all(y.index.levels[2] == index.levels[1])
     np.testing.assert_array_equal(y.values, y_expected)
+
+    # non-existing folder
+    with pytest.raises(FileNotFoundError):
+        feature.process_folder('bad-folder')
+
+    # empty folder
+    root = str(tmpdir.mkdir('empty'))
+    y = feature.process_folder(root)
+    pd.testing.assert_frame_equal(
+        y,
+        pd.DataFrame(
+            dtype=object,
+            columns=feature.column_names,
+        ),
+    )
 
 
 def test_process_func_args():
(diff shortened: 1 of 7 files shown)


tests/test_feature.py
```diff
@@ -286,6 +287,21 @@ def test_process_folder(tmpdir):
...
+    with pytest.raises(FileNotFoundError):
+        feature.process_folder('bad-folder')
+
+    # empty folder
+    root = str(tmpdir.mkdir('empty'))
+    y = feature.process_folder(root)
+    pd.testing.assert_frame_equal(
+        y,
+        pd.DataFrame(
+            dtype=object,
```
frankenjoe, line 300: Strange we need `object` here, I was expecting `float`...

hagenw: I get:

```python
>>> pd.Series()
<ipython-input-101-fa15c544da3b>:1: DeprecationWarning: The default dtype for empty Series will be 'object' instead of 'float64' in a future version. Specify a dtype explicitly to silence this warning.
  pd.Series()
Out[101]: Series([], dtype: float64)
>>> pd.Series(dtype=object)
Series([], dtype: object)
>>> pd.Series(dtype=float)
Series([], dtype: float64)
```
So, I guess there is something wrong (or is it a feature?) when converting the series to a dataframe.
If I do
```python
print(y)
print(y.dtypes)
```
I get
```
Empty DataFrame
Columns: [o1-0, o2-0, o3-0, o1-1, o2-1, o3-1]
Index: []
o1-0    object
o2-0    object
o3-0    object
o1-1    object
o2-1    object
o3-1    object
dtype: object
```

tests/test_feature.py
```diff
@@ -286,6 +287,21 @@ def test_process_folder(tmpdir):
...
 
+    # non-existing folder
+    with pytest.raises(FileNotFoundError):
+        feature.process_folder('bad-folder')
+
+    # empty folder
+    root = str(tmpdir.mkdir('empty'))
+    y = feature.process_folder(root)
+    pd.testing.assert_frame_equal(
+        y,
```
hagenw, line 298: We have a dataframe here:

```suggestion
    df = feature.process_folder(root)
    pd.testing.assert_frame_equal(
        df,
```

tests/test_segment.py
```diff
@@ -99,11 +99,20 @@ def test_folder(tmpdir, num_workers, multiprocessing):
...
     assert all(result.levels[2] == INDEX.levels[1])
 
+    # non-existing folder
+    with pytest.raises(FileNotFoundError):
+        segment.process_folder('bad-folder')
+
+    # empty folder
+    root = str(tmpdir.mkdir('empty'))
+    y = segment.process_folder(root)
+    pd.testing.assert_index_equal(y, audformat.filewise_index())
```
hagenw, line 114: And here we have an index:
```suggestion
    index = segment.process_folder(root)
    pd.testing.assert_index_equal(index, audformat.filewise_index())
```

hagenw (commented):
Good, that you found that error. I'm using `process_folder()` very rarely, so I guess I would never have found this error ;)

tests/test_feature.py
hagenw: >  So it seems they will only change the default type to float for Series in the future.

No, they will change it to `object`:

> The default dtype for empty Series will be 'object' 

Do you think we should do the same then?

frankenjoe: Ah, I see. So they actually change the default behavior of `Series` to that of `DataFrame`. That makes sense of course!

> Do you think we should do the same then?

Mhh, good question. Probably it makes indeed sense to go with `object`, since this is most generic data type.

frankenjoe: I created https://github.com/audeering/audinterface/issues/35

hagenw: Why not directly change it here, we have to break one behavior anyway, at the moment you decided to change the dataframe to `float`. What about other places where we return an empty dataframe?

frankenjoe: > Why not directly change it here

So you mean changing to `object` when we return an empty `Series` and do not return `DataFrame` as `float`, correct?

> What about other places where we return an empty dataframe?

As far as I saw there's only one place, but I will check again.

hagenw: > So you mean changing to object when we return an empty Series and do not return DataFrame as float, correct?

Yes, exactly.

frankenjoe: Done, this basically reverts some of the changes from #32

PR #36, opened 2022-02-02
Feature: support plain string for single feature

frankenjoe:
If a single string is passed to `Feature' it will be automatically converted to an array.


hagenw:
Great.

PR #39, opened 2022-02-07
Release 0.8.1

hagenw:
image


hagenw:
Need to fix the docs before with https://github.com/audeering/audinterface/pull/40

frankenjoe:
I guess #40 is a too small change that we want to mention in the changelog, right?

hagenw:
Yes, it's nothing the user has to care about.

PR #43, opened 2022-06-02
Add cache option to Feature.process_index()

### LINKED ISSUE #42: Add option to cache results with process_index()
Add option to pass a cache directory to `process_index()`. Use `hash = audformat.utils.hash(index)` to create file name or load from cache if file exists.

Possible extension: combine with a hash that is generated from the `process_func()` to avoid accidentally loading results generated with a different function.

frankenjoe:
Closes #42 

image

### Example

```python
import audb
import audeer
import audinterface


db = audb.load(
    'emodb',
    version='1.2.0',
    format='wav',
    mixdown=True,
    sampling_rate=16000,
)

def process_func(x, sr):
    return [x.mean(), x.min(), x.max()]

    ['mean', 'min', 'max'],
    process_func=process_func,
)
feature.process_index(
    db.files,
    cache_root='cache',
)
```
```
                                                                                         mean  ...       max
file                                               start  end                                  ...          
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:01.898250    -0.000311  ...  0.999969
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:01.611250    -0.000312  ...  0.935303
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:01.877812500 -0.000296  ...  0.975800
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:02.006250    -0.000171  ...  0.999634
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:01.439812500 -0.000331  ...  0.889465
...                                                                                       ...  ...       ...
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:03.442687500 -0.000268  ...  0.788452
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:03.500625    -0.001218  ...  0.999969
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:03.934187500 -0.001157  ...  0.999969
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:02.414125    -0.000195  ...  0.999969
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:02.522499999 -0.000464  ...  0.999908

[535 rows x 3 columns]
```

```python
cache_path = audeer.path('cache', f'{audformat.utils.hash(db.files)}.pkl')
pd.read_pickle(cache_path)
```
```
                                                                                         mean  ...       max
file                                               start  end                                  ...          
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:01.898250    -0.000311  ...  0.999969
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:01.611250    -0.000312  ...  0.935303
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:01.877812500 -0.000296  ...  0.975800
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:02.006250    -0.000171  ...  0.999634
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:01.439812500 -0.000331  ...  0.889465
...                                                                                       ...  ...       ...
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:03.442687500 -0.000268  ...  0.788452
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:03.500625    -0.001218  ...  0.999969
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:03.934187500 -0.001157  ...  0.999969
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:02.414125    -0.000195  ...  0.999969
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 0 days 00:00:02.522499999 -0.000464  ...  0.999908

[535 rows x 3 columns]
```

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -301,12 +301,16 @@ class Feature:
             index: pd.Index,
             *,
             root: str = None,
+            cache_root: str = None,
     ) -> pd.DataFrame:
         r"""Extract features from an index conform to audformat_.
 
         Args:
             index: index with segment information
             root: root folder to expand relative file paths
+            cache_root: cache result under this folder.
+                Filename is created with
+                :func:`audformat.utils.hash`
 
         Raises:
             RuntimeError: if sampling rates do not match
@@ -314,12 +318,41 @@ class Feature:
             RuntimeError: if multiple frames are returned,
                 but ``win_dur`` is not set
             ValueError: if index is not conform to audformat_
+            ValueError: if cached file has different column names
 
         .. _audformat: https://audeering.github.io/audformat/data-format.html
 
         """
-        series = self.process.process_index(index, root=root)
-        return self._series_to_frame(series)
+        cache_path = None
+
+        if cache_root is not None:
+            cache_root = audeer.mkdir(cache_root)
+            hash = audformat.utils.hash(index)
+            cache_path = os.path.join(cache_root, f'{hash}.pkl')
+
+        if cache_path and os.path.exists(cache_path):
+
+            df = pd.read_pickle(cache_path)
+            if list(df.columns) != self.column_names:
+                raise ValueError(
+                    f'Found cached file with different column names: '
+                    f'{list(df.columns)} '
+                    f'!= '
+                    f'{self.column_names}'
+                )
+
+        else:
+
+            y = self.process.process_index(
+                index,
+                root=root,
+            )
+            df = self._series_to_frame(y)
+
+            if cache_path is not None:
+                df.to_pickle(cache_path, protocol=4)
+
+        return df
 
     def process_signal(
             self,
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -309,7 +309,7 @@ class Feature:
             index: index with segment information
             root: root folder to expand relative file paths
             cache_root: cache result under this folder.
-                Filename is created with
+                Filename will be created with
                 :func:`audformat.utils.hash`
 
         Raises:
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,15 +19,15 @@ classifiers =
     Operating System :: OS Independent
     Programming Language :: Python
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.7 
-    Programming Language :: Python :: 3.8 
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
     Topic :: Scientific/Engineering
 
 [options]
 packages = find:
 install_requires =
-    audformat >=0.10.1,<2.0.0
+    audformat >=0.12.1,<2.0.0
     audresample >=1.1.0,<2.0.0
 setup_requires =
     setuptools_scm
--- a/tests/test_feature.py
+++ b/tests/test_feature.py
@@ -759,6 +759,8 @@ def test_process_signal_from_index(index, expected_features):
 
 def test_process_index(tmpdir):
 
+    cache_root = os.path.join(tmpdir, 'cache')
+
     feature = audinterface.Feature(
         feature_names=('o1', 'o2', 'o3'),
         process_func=feature_extractor,
@@ -787,6 +789,7 @@ def test_process_index(tmpdir):
     assert y.index.get_level_values('file')[0] == path
     np.testing.assert_array_equal(y.values, y_expected)
     assert y.columns.tolist() == feature.column_names
+    y = feature.process_index(index)
 
     # relative paths
     index = audformat.segmented_index([file] * 2, [0, 1], [2, 3])
@@ -794,6 +797,44 @@ def test_process_index(tmpdir):
     assert y.index.get_level_values('file')[0] == file
     np.testing.assert_array_equal(y.values, y_expected)
     assert y.columns.tolist() == feature.column_names
+
+    # cache result
+    y = feature.process_index(
+        index,
+        root=root,
+        cache_root=cache_root,
+    )
+    os.remove(path)
+
+    # fails because file does not exist
+    with pytest.raises(RuntimeError):
+        feature.process_index(
+            index,
+            root=root,
+        )
+
+    # loading from cache still works
+    y_cached = feature.process_index(
+        index,
+        root=root,
+        cache_root=cache_root,
+    )
+    pd.testing.assert_frame_equal(y, y_cached)
+
+    # interface with different feature names
+    feature = audinterface.Feature(
+        feature_names=('some', 'other', 'names'),
+        process_func=feature_extractor,
+        channels=range(NUM_CHANNELS),
+    )
+
+    # fails because column names do not match
+    with pytest.raises(ValueError):
+        feature.process_index(
+            index,
+            root=root,
+            cache_root=cache_root,
+        )
 
 
 @pytest.mark.parametrize(


audinterface/core/feature.py
```diff
@@ -301,25 +301,58 @@ def process_index(
...
+
+        if cache_root is not None:
+            cache_root = audeer.mkdir(cache_root)
+            hash = audformat.utils.hash(index)
+            cache_path = os.path.join(cache_root, f'{hash}.pkl')
+
+        if cache_path and os.path.exists(cache_path):
+
+            df = pd.read_pickle(cache_path)
+            if list(df.columns) != self.column_names:
```
frankenjoe, line 336: There is of course still the risk that you run a different model with the same output names. In that case a user would not notice. To make it safer we could process the first segment of the index and compare it with the cached values.

frankenjoe: I actually decided to implement it that way. Now not only the columns have to match, but also the values of the first segment. It's a bit safer that way.

hagenw: The downside is that you need to be able to run the model, even when loading from cache. E.g. if the model can only be run on GPU and no GPU is available, it would be more convenient if we don't have to process the first segment.

I think it would be safer if we could achieve something like `<cache-root>/<model>/<hash>.pkl`. The question is then how to get `<model>`. One way is that a user might be providing it by an attribute or an argument when instantiating the `audinterafce` object. We might fall back to a standard name if we cannot get it, but it would still be nice, if every model that is loaded with `audmodel` or `audonnx` could automatically result in some model name.

frankenjoe: > I think it would be safer if we could achieve something like `<cache-root>/<model>/<hash>.pkl`. The question is then how to get `<model>`. One way is that a user might be providing it by an attribute or an argument when instantiating the `audinterafce` object. We might fall back to a standard name if we cannot get it, but it would still be nice, if every model that is loaded with `audmodel` or `audonnx` could automatically result in some model name.

But `audinterface` is meant to work with any processing function not just `audonnx` models.

frankenjoe:
> The downside is that you need to be able to run the model, even when loading from cache. E.g. if the model can only be run on GPU and no GPU is available, it would be more convenient if we don't have to process the first segment.

I think we should always assume that the processing function is working, as we cannot know in advance if we can load from cache or not. Otherwise, it makes no sense to create the interface and the user can simply directly load the cached file.

hagenw:
> and the user can simply directly load the cached file.

No, because usually there is no easy way to figure out where it is stored as you need to know the correct hash. So, usually it is much more convincing to just rerun the process function.

audinterface/core/feature.py
hagenw: I know, in that case we can still use a fallback name for `<model>`.

hagenw: > There is of course still the risk that you have a different process function but use the same feature names

This will happen all the time. If you would like to compare different gender models, you would always call the feature gender. Or even don't use `audinterface.Feature` but `audinterface.Process` to get the values back as a Series.

frankenjoe:
> No, because usually there is no easy way to figure out where it is stored as you need to know the correct hash.

In the docstring we explain it:

image

>  So, usually it is much more convincing to just rerun the process function.

Which is now problem if it is a working processing function :) Just in the case where you cannot create the interface at all you can still manually load the cached file.

But ok, I'm also fine with removing the additional check and we simply leave it to the user to not use the wrong cache folder.

audinterface/core/feature.py
frankenjoe: Mhh, not convinced :) Then I would rather skip the additional test.

hagenw: I would also skip the additional test, but I would like to have the possibility to specify a model name. As I don't like calling my features `gender-model1`, `gender-model2`, ...

frankenjoe: >  The question is then how to get <model>. One way is that a user might be providing it by an attribute or an argument when instantiating the audinterafce object. 

Then we again have the problem that the user accidentally provides the wrong name and will not notice that the wrong cache file is loaded.

frankenjoe: > I would also skip the additional test, but I would like to have the possibility to specify a model name. As I don't like calling my features gender-model1, gender-model2, ...

Sorry, but I think I don't get your point.

hagenw: Yes, we cannot avoid this.

But that is the reason why I proposed to set the name automatically, when using `audmodel` or `audonnx`.

Another advantage of not storing everything in the same folder is that it will become slightly easier to clean up the cache after a while.

frankenjoe: But you don't have to store it in the same folder. The way I see it is that with `model1` you store it in `/cache/model1` and with `model2` in `/cache/model2` etc.

frankenjoe: I just don't think we need to introduce another argument for that.

frankenjoe: And as I said the risk of loading the wrong cache file still exists, so it doesn't helper here either.

hagenw: > But you don't have to store it in the same folder. The way I see it is that with model1 you store it in /cache/model1 and with model2 in /cache/model2 etc.

> I just don't think we need to introduce another argument for that.

I agree.

I also misunderstood the `cache_root` argument. So at the moment we do not support using a shared cache folder automatically (in that case it would be handy to automatically get model name from `process_func.uid`), but the user has to provide the `cache_root` argument explicitly, otherwise no cache is used at all.

Maybe this is indeed the best way to start and see if it works out.
But I would then use only the index and not the feature name for hash calculation, then you can also use the cache with `audinterface.Process`.

frankenjoe: > But I would then use only the index and not the feature name for hash calculation, then you can also use the cache with audinterface.Process.

But that's what we already do, we only use the index to create the hash.

audinterface/core/feature.py
```diff
@@ -301,25 +301,86 @@ def process_index(
...
+        else:
+
+            y = self.process.process_index(
+                index,
+                root=root,
+            )
+            df = self._series_to_frame(y)
+
+            if cache_path is not None:
+                df.to_pickle(cache_path, protocol=4)
```
hagenw, line 381: As we call `Process.process_index()` inside `Feature.process_index()` anyway, we could also add the caching directly to `audinterface.Process` and simply hand the `cache_root` when calling `self.process.process_index()`.

The advantage would be that we would then support `Process` and `Feature` with the same cache folder.
The downside is that we would need to call `self._series_to_frame(y)` for `Feature` every time we load from cache.
But in my opinion the advantage of supporting both would be bigger.

frankenjoe: Actually, that's how I implemented it in the first place, but as you said there is the downside that we need to call `self._series_to_frame(y)`. The other disadvantage is that a user then cannot directly load the `DataFrame` from the cache. And personally, I use `Feature` in 99% of the times and almost never `Process`. So if we want to have the cache feature with `Process` I'd suggest we add it separately.

hagenw: >  I use Feature in 99% of the times and almost never Process

For me it's more the opposite as `Process` allows you:

```python
df['prediction'] = model.process_index(df.index)
```

But it doesn't matter, I would simply say we should support the cache in both.

> So if we want to have the cache feature with Process I'd suggest we add it separately.

Yes, but they will try to load the same PKL file as the file name is only given by the index. Will this also work if we go with your solution here and then later on add the same to `Process`?

hagenw: > The other disadvantage is that a user then cannot directly load the DataFrame from the cache

It's still a valid PKL file and the user can load it. It will just contain the raw predictions of the model, but I don't see anything wrong with this.

frankenjoe: > It's still a valid PKL file and the user can load it. It will just contain the raw predictions of the model, but I don't see anything wrong with this.

1. returns a `pd.Series` instead of a `pd.DataFrame`
2. does not contain feature names
3. in case of framewise output there is no correct time information for the frames

frankenjoe: > Yes, but they will try to load the same PKL file as the file name is only given by the index.

You have to use a different cache folder. But you should anyway use a different one for every interface.

audinterface/core/feature.py
frankenjoe: So would you suggest to remove both security checks then?

frankenjoe: I removed the sanity checks.

audinterface/core/feature.py
frankenjoe: I removed all sanity checks. And as you said, the caching functionality we add here is not meant to be used as a shared cache (therefore we have `audfeatures`). And also the user is responsible to take care of not loading a cache file generated with a different processing function.

audinterface/core/feature.py
hagenw: I think it makes indeed sense to remove the sanity check if the output of the model matches. The oterh one I will discuss below.

audinterface/core/feature.py
frankenjoe: I just removed the check :)

And I would certainly not replace the feature names silently, as it's a strong sign that the values are also not the desired once.

Frankly, I would simply stay with the solution now that we have no checks.

hagenw: If you don't like renaming the feature column, you could still raise an error as before, but then the error message should list how the expected feature names are.

frankenjoe: The main danger I see is that you have model A and model B, which both output the same classes and that you accidentally use the same cache folder. In that case, comparing the number of columns and feature names will not help. That's why I had initially added the second check that the values of the first segment have to match. So I would either add both checks or no checks at all.

hagenw: If we stay with the current solution, then we might get the following problem.

Let's assume we start with the following code

```python
feature = 'gender'
interface = audinterface.Feature([feature])
df = interface.process_index(index, cache_root='~/cache/model_id')
df[feature]
```

Now at some point we decide the feature should be called sex instead of gender and modify the code to

```python
feature = 'sex'
interface = audinterface.Feature([feature])
df = interface.process_index(index, cache_root='~/cache/model_id')
df[feature]
```

This will simply fail and from just reading the code it will be hard to find the error.

frankenjoe: If we decide to also support caching to `Process` we also cannot check for columns.

audinterface/core/feature.py
```diff
@@ -301,12 +301,22 @@ def process_index(
             index: pd.Index,
             *,
             root: str = None,
+            cache_root: str = None,
     ) -> pd.DataFrame:
         r"""Extract features from an index conform to audformat_.
 
+        If cache folder is not ``None``,
```
hagenw, line 308:
```suggestion
        If ``cache_root`` is not ``None``,
```

PR #45, opened 2022-06-08
Add cache option to Process.process_index()

### LINKED ISSUE #42: Add option to cache results with process_index()

hagenw:
Closes #42 

Similar to https://github.com/audeering/audinterface/pull/43 this adds the option to cache results when running `audinterface.Process.process_index()`.

image

### DIFF
--- a/audinterface/core/process.py
+++ b/audinterface/core/process.py
@@ -322,12 +322,22 @@ class Process:
             index: pd.Index,
             *,
             root: str = None,
+            cache_root: str = None,
     ) -> pd.Series:
         r"""Process from an index conform to audformat_.
+
+        If ``cache_root`` is not ``None``,
+        a hash value is created from the index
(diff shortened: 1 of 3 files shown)


frankenjoe:
Great! Could you please do me an favor and in `feature.py` change:

```python
        if cache_path and os.path.exists(cache_path):
            return pd.read_pickle(cache_path)

        else:
```



```python
        if cache_path and os.path.exists(cache_path):
            df = pd.read_pickle(cache_path)
        else:
```

frankenjoe:
What's weird is that I didn't get an email notification about this PR...

PR #46, opened 2022-06-09
DOC: add usage section

### LINKED ISSUE #5: Add usage section to documentation
We don't have a usage section at the moment, which is not very nice.

I also think we need two different sections. One, how to use `audinterface` to turn your class (e.g. model) into an application that can be applied to files, etc. And a second one, showing how you can use `audinterface` when developing an application such as `opensmile`.

hagenw:
Closes #5.

This adds a first basic usage section to the documentation, covering:

* using `audinterface.Process`
* using `audinterface.Feature`
* building a feature extraction object with `audinterface.Feature` and `audobject.Object`

Building the documentation will fail as I get the following warning due to storing and loading the feature extraction object:

```
WARNING: Cell printed to stderr:
/home/audeering.local/hwierstorf/.envs/audinterface/lib/python3.8/site-packages/audobject/core/utils.py:51: RuntimeWarning: Could not determine a version for module '__main__'.
  warnings.warn(
```

Any idea how to avoid this?

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -247,7 +247,7 @@ Segmentation interface
         return index
 
     interface = audinterface.Segment(process_func=segments)
-    idx = interafce.process_file(files[0])
+    idx = interface.process_file(files[0])
     idx
 
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -215,6 +215,12 @@ and re-instantiate it from there.
 Segmentation interface
 ----------------------
 
+When the result of the processing function is an index
+it is recommended to use :class:`audinterface.Segment`,
+which returns a segmented index conform to audformat_.
+An example for such a processing function
+would be a voice activity detection algorithm.
+
 .. jupyter-execute::
 
     import auditok
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -112,6 +112,49 @@ and assigns names to single features.
     df
 
 
+Create a framewise feature extractor
+------------------------------------
+
+If the feature extractor does not return
+one set of features for the whole signal,
+but does return features
+in a framewise manner,
+you should specify the ``win_dur``
+and ``hop_dur`` arguments
+of :class:`audinterface.Feature`.
+
+.. jupyter-execute::
+
+    def features(signal, sampling_rate, win_dur, hop_dur):
+        win_dur = int(win_dur * sampling_rate)
+        hop_dur = int(hop_dur * sampling_rate)
+        dur = signal.shape[1]
+        starts = np.arange(0, dur - win_dur, hop_dur)
+        ends = np.arange(win_dur, dur, hop_dur)
+        mean = np.array(
+            [
+                [signal[:, s:e].mean()]
+                for s, e in zip(starts, ends)
+            ]
+        )
+        return mean.T  # return with shape (num_features, num_frames)
+
+    win_dur = 0.2
+    hop_dur = 0.1
+    interface = audinterface.Feature(
+        ['mean'],
+        process_func=features,
+        process_func_args={
+            'win_dur': win_dur,
+            'hop_dur': hop_dur,
+        },
+        win_dur=win_dur,
+        hop_dur=hop_dur,
+    )
+    df = interface.process_index(index)
+    df
+
+
 Create a serializable feature extractor
 ---------------------------------------
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -33,7 +33,7 @@ to audio files.
 The only prerequisite is
 the algorithm provides a callable
 that takes at least the signal
-as a :class:`np.ndarray`
+as a :class:`numpy.ndarray`
 and the sampling rate as input.
 
 The interface can then apply the algorithm
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -12,7 +12,12 @@
         return df._repr_html_()
 
 
+    def index_to_html(self):
+        return self.to_frame(index=False)._repr_html_()
+
+
     setattr(pd.Series, '_repr_html_', series_to_html)
+    setattr(pd.Index, '_repr_html_', index_to_html)
     pd.set_option('display.max_rows', 6)
     pd.set_option('display.max_columns', 3)
 
@@ -91,7 +96,7 @@ as a :class:`pandas.Series`.
     y
 
 
-Feature extractor
+Feature interafce
 -----------------
 
 When the result of the processing function has multiple dimensions
@@ -113,17 +118,17 @@ and assigns names to the dimensions/features.
     df
 
 
-Framewise feature extractor
+Framewise feature interafce
 ---------------------------
 
-If the feature extractor does not return
+If a processing function does not return
 one set of features for the whole signal,
 but does return features
 in a framewise manner,
 you should specify the ``win_dur``
 and ``hop_dur`` arguments
 of :class:`audinterface.Feature`.
-It's also important the feature extraction function
+It's also important the processing function
 returns the value in the correct shape,
 namely ``(num_channels, num_features, num_frames)``,
 whereas the first dimension is optionally.
@@ -162,7 +167,7 @@ whereas the first dimension is optionally.
     df
 
 
-Serializable feature extractor
+Serializable feature interafce
 ------------------------------
 
 To use a feature extractor as an input transform
@@ -205,6 +210,46 @@ and re-instantiate it from there.
     fex2 = audobject.from_yaml('mean-std.yaml')
     df = fex2.process_index(index)
     df
+
+
+Segmentation interface
+----------------------
+
+.. jupyter-execute::
+
+    import auditok
+
+    def segments(signal, sampling_rate):
+
+        # Convert floating point array to 16bit PCM little-endian
+        ints = (signal[0, :] * 32767).astype(np.int16)
+        little_endian = ints.astype('<u2')
+        signal = little_endian.tobytes()
+
+        regions = auditok.split(
+            signal,
+            sampling_rate=sampling_rate,
+            sample_width=2,
+            channels=1,
+            min_dur=0.2,
+            energy_threshold=70,
+        )
+        index = pd.MultiIndex.from_tuples(
+            [
+                (
+                    pd.Timedelta(region.meta.start, unit='s'),
+                    pd.Timedelta(region.meta.end, unit='s'),
+                )
+                for region in regions
+            ],
+            names=['start', 'end'],
+        )
+        return index
+
+    interafce = audinterface.Segment(process_func=segments)
+    idx = interafce.process_file(files[0])
+    idx
+
 
 .. _audformat: https://audeering.github.io/audformat/
 .. _emodb: http://emodb.bilderbar.info
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -28,7 +28,7 @@ Usage
 The basic idea of :mod:`audinterface` is
 to provide easy and standardized interfaces
 to apply a machine learning model,
-or other digital signal processing algorithms (DSP)
+or other digital signal processing algorithms
 to audio files.
 The only prerequisite is
 the algorithm provides a callable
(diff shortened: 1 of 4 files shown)


frankenjoe:
> Any idea how to avoid this?

did you try to set `__version__ = '1.0.0' ` at the top the file?

hagenw:
> > Any idea how to avoid this?
> 
> did you try to set `__version__ = '1.0.0' ` at the top the file?

Thanks, this works indeed.

docs/usage.rst
```diff
@@ -0,0 +1,156 @@
...
+
+    media = ['wav/03a01Fa.wav', 'wav/03a01Nc.wav', 'wav/03a01Wa.wav']
+    db = audb.load('emodb', version='1.2.0', media=media, verbose=False)
+
+    files = list(db.files)
+    folder = os.path.dirname(files[0])
+    index = db['emotion'].index
+
+
+DSP algorithm returning a series
```
hagenw, line 62: I call it here "DSP algorithm" as my original goal was to also add a section for an example using a "Machine learning model". But I'm not sure if this is needed and if we have a small one that we use here.

frankenjoe: I am not sure if really need that, but if we want to, we can create a dummy model with `audonnx.testing`.

docs/usage.rst
```diff
@@ -0,0 +1,162 @@
...
+
+The basic idea of :mod:`audinterface` is
+to provide easy and standardized interfaces
+to apply a machine learning model,
+or other digital signal processing algorithms (DSP)
+to audio files.
+The only prerequisite is
+the algorithm provides a callable
+that takes at least the signal
+as a :class:`np.ndarray`
```
frankenjoe, line 36:
```suggestion
as a :class:`numpy.ndarray`
```

docs/usage.rst
frankenjoe: Ok, then I suggest we also choose a more generic name for the section, maybe: "Create a processing interface"

docs/usage.rst
```diff
@@ -0,0 +1,162 @@
...
+    y
+
+
+DSP algorithm returning a dataframe
+-----------------------------------
+
+When using a DSP algorithm as feature extractor,
+it is recommended to use :class:`audinterface.Feature`,
+which returns results as a :class:`pd.DataFrame`
+and assigns names to the single features.
```
frankenjoe, line 99: Another advantage is that we can set a window and hop duration in case of features with dynamic length. I think we should highlight by adding a second example.

docs/usage.rst
```diff
@@ -0,0 +1,162 @@
...
+    interface = audinterface.Feature(
+        ['mean', 'std'],
+        process_func=features,
+    )
+
+    df = interface.process_index(index)
+    df
+
+
+Creating a feature extraction object
```
frankenjoe, line 115: Maybe: Create a serializable feature extractor?

docs/usage.rst
```diff
@@ -0,0 +1,162 @@
...
+    df = fex.process_index(index)
+    df
+
+The advantage of the feature extraction object is
+that we can save it to a YAML file
+and re-instantiate it from there.
+
+.. jupyter-execute::
+
+    fex.to_yaml('mean-std.yml')
```
frankenjoe, line 154:
```suggestion
    fex.to_yaml('mean-std.yaml')
```

docs/usage.rst
```diff
@@ -0,0 +1,162 @@
...
+    fex2 = audobject.from_yaml('mean-std.yml')
```
frankenjoe, line 155:
```suggestion
    fex2 = audobject.from_yaml('mean-std.yaml')
```

we usually use "yaml" (in `audonnx` we even force it), so I think it's better to always stick to `.yaml`.

frankenjoe:
Good to see we will finally have a usage section!

docs/usage.rst
hagenw: I added such an example under the new subsection "Create a framewise feature extractor".
But maybe my implementation of calculating the mean on a framewise basis is slightly clumsy ;)

image

docs/usage.rst
hagenw: Changed to "Create a serializable feature extractor"

hagenw:
At the moment every subsection starts with "Create a"

image

should I maybe change to

image

docs/usage.rst
frankenjoe: Maybe:

```python
def features(signal, sampling_rate, win_dur, hop_dur):
    win_dur = int(win_dur * sampling_rate)
    hop_dur = int(hop_dur * sampling_rate)
    starts = np.arange(0, signal.shape[1] - win_dur, hop_dur)
    mean = [[signal[:, s:s+win_dur].mean() for s in starts]]
    return mean
```

docs/usage.rst
frankenjoe: Yes, makes sense.

docs/usage.rst
```diff
@@ -0,0 +1,209 @@
...
+    y = interface.process_files(files)
+    y = interface.process_folder(folder)
+    y = interface.process_index(index)
+    y
+
+
+Create a feature extractor
+--------------------------
+
+When using an algorithm as feature extractor,
```
frankenjoe, line 96: Maybe we should be a little bit more generic here. E.g.: When the result of the processing function has multiple dimensions it is recommended to use :class:`audinterface.Feature`, which returns a :class:`pandas.DataFrame`
with named dimensions.

docs/usage.rst
frankenjoe: Or we switch to an example where we interface an external library, e.g.:

```python
win_dur = 0.02
hop_dur = 0.01
n_mfcc = 13
    [f'mfcc-{idx}' for idx in range(n_mfcc)],
    process_func=features,
    process_func_args={
        'win_dur': win_dur,
        'hop_dur': hop_dur,
        'n_mfcc': n_mfcc,
    },
    hop_dur=hop_dur,
    win_dur=win_dur,
)
df = interface.process_files(db.files[:10])
df
```
```
                                                                                                      mfcc-0  ...    mfcc-12
file                                               start                  end                                 ...           
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 00:00:00        0 days 00:00:00.020000 -611.993286  ...   1.151396
                                                   0 days 00:00:00.010000 0 days 00:00:00.030000 -668.175842  ...  14.068544
                                                   0 days 00:00:00.020000 0 days 00:00:00.040000 -664.612793  ...   7.949759
                                                   0 days 00:00:00.030000 0 days 00:00:00.050000 -667.714722  ...  12.957479
                                                   0 days 00:00:00.040000 0 days 00:00:00.060000 -669.367432  ...   4.396849
...                                                                                                      ...  ...        ...
/media/jwagner/Data/audb/emodb/1.2.0/fe182b91/w... 0 days 00:00:01.650000 0 days 00:00:01.670000 -619.806335  ...  11.462805
                                                   0 days 00:00:01.660000 0 days 00:00:01.680000 -622.713928  ...  13.530191
                                                   0 days 00:00:01.670000 0 days 00:00:01.690000 -624.977356  ...  12.032556
                                                   0 days 00:00:01.680000 0 days 00:00:01.700000 -621.033264  ...  11.019306
                                                   0 days 00:00:01.690000 0 days 00:00:01.710000 -622.040344  ...  11.202772

[1743 rows x 13 columns]
```

docs/usage.rst
```diff
@@ -0,0 +1,209 @@
...
+a folder,
+and an index.
+
+.. jupyter-execute::
+
+    import audb
+    import os
+
+    media = ['wav/03a01Fa.wav', 'wav/03a01Nc.wav', 'wav/03a01Wa.wav']
+    db = audb.load('emodb', version='1.2.0', media=media, verbose=False)
```
frankenjoe, line 55: I'd recommend we set `full_path=False` and then use `process_index(..., root=db.root)` to avoid showing the full path in the output.

docs/usage.rst
frankenjoe: I think it's general a good idea to cover this kind of use-case.

docs/usage.rst
hagenw: Good to know, maybe we should then also extend the docstring and be more precise there.
I thought when using `root=` it will also add this to the path in the returned index.

But if not, it is indeed much more elegant than my current solution.

hagenw: > it is indeed much more elegant than my current solution.

I just realized, this refers more to what I did in
https://github.com/audeering/audinterface/pull/48,
where I used `audformat.utils.map_file_path()` to remove the absolute path in the output.

Here, in the usage documentation I think it makes more sense to stay with `full_path=True` as this is the behavior the user would most likely need. The resulting dataframes will look ok in the HTML version of the docs, as the full path is usually shorter on the Github runner as it is when building locally.

frankenjoe: ok

docs/usage.rst
hagenw: Changed accordingly.

hagenw:
> Should we also add an example for `Segment`?

I added one, but it was not so easy to find a processing function that returns segments.
For now I ended up using https://auditok.readthedocs.io/

hagenw:
> At the moment every subsection starts with "Create a"
> 
> image
> 
> should I maybe change to
> 
> image

I adjusted the headings accordingly, and also changed the last word always to "interface"

docs/usage.rst
hagenw: Good idea, I switched to `librosa.feature.mfcc`

frankenjoe:
> For now I ended up using https://auditok.readthedocs.io/

Interesting, they even offer real-time detection. Could this also be a candidate benchmark against?

hagenw:
> > For now I ended up using https://auditok.readthedocs.io/
> 
> Interesting, they even offer real-time detection. Could this also be a candidate benchmark against?

Maybe, but at least according to this it performs not as good as others: https://github.com/wiseman/py-webrtcvad/issues/68

docs/usage.rst
```diff
@@ -0,0 +1,257 @@
...
+                    pd.Timedelta(region.meta.start, unit='s'),
+                    pd.Timedelta(region.meta.end, unit='s'),
+                )
+                for region in regions
+            ],
+            names=['start', 'end'],
+        )
+        return index
+
+    interafce = audinterface.Segment(process_func=segments)
```
frankenjoe, line 249:
```suggestion
    interface = audinterface.Segment(process_func=segments)
```

docs/usage.rst
```diff
@@ -0,0 +1,257 @@
...
+    idx = interafce.process_file(files[0])
```
frankenjoe, line 250:
```suggestion
    idx = interface.process_file(files[0])
```

frankenjoe: I think I would also rename to `index`, with `idx` I associate an integer.

frankenjoe: Ah wait, we already have a variable `index`.

frankenjoe: Actually, is there a specific reason that you assign the result and not just show the output directly (applies also to the other examples).

hagenw: The reason behind is that the user can more easily see what type the method returns.
If we just show the HTML formated output I'm afraid that it is not that easy to differentiate between Series, DataFrame, and Index

frankenjoe: ok

hagenw: And yes, I used `idx` instead of `index` to not overwrite the already assigned `index` variable.

We could call it `segmented_index`, but I was also not convinced to add an additional word as we also don't do it for the other examples.

frankenjoe: Yes, I guess there is no perfect solution, so let's stick with `idx`.

docs/usage.rst
```diff
@@ -0,0 +1,257 @@
...
+
+    fex.to_yaml('mean-std.yaml')
+    fex2 = audobject.from_yaml('mean-std.yaml')
+    df = fex2.process_index(index)
+    df
+
+
+Segmentation interface
+----------------------
+
```
frankenjoe, line 217: Like with the other sections I think we should provide a short introduction here.

hagenw: Oh, totally forgot that.

I now added

image

frankenjoe: Cool, thanks. No more comments, so I think we are good to go.

PR #48, opened 2022-06-09
DOC: add docstring examples

### LINKED ISSUE #11: Add docstring examples
We currently do not provide code examples in the docstrings.

hagenw:
Closes #11 

This adds examples to all API functions and classes.
In addition, it removes the example from `__init__.py` as we now have the same in the docstrings.

### audinterface.Feature

image

### audinterafce.Process

image

### audinterface.ProcessWithContext

image

### audinterface.Segment

image

### audinterface.utils.read_audio

image

### DIFF
--- a/audinterface/core/segment.py
+++ b/audinterface/core/segment.py
@@ -156,24 +156,27 @@ class Segment:
                     ('0 days 00:00:00.700000', '0 days 00:00:00.900000')],
                    names=['start', 'end'])
         >>> import audb
-        >>> import audformat
-        >>> db = audb.load('emodb', version='1.2.0', media='wav/03a01Fa.wav', verbose=False)
+        >>> db = audb.load(
+        ...     'emodb',
+        ...     version='1.2.0',
+        ...     media='wav/03a01Fa.wav',
+        ...     full_path=False,
+        ...     verbose=False,
+        ... )
         >>> interface = Segment(
         ...     process_func=segment,
         ...     process_func_args={'win_size': 0.5, 'hop_size': 0.25},
         ... )
-        >>> index = interface.process_index(db['emotion'].index)
-        >>> index = audformat.utils.map_file_path(index, os.path.basename)
-        >>> index
-        MultiIndex([('03a01Fa.wav',        '0 days 00:00:00', ...),
-                    ('03a01Fa.wav', '0 days 00:00:00.250000', ...),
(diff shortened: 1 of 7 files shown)


audinterface/core/utils.py
```diff
@@ -123,7 +123,19 @@ def read_audio(
...
-    """
+    Example:
+        >>> import audb
+        >>> media = audb.load_media(
+        ...     'emodb',
+        ...     'wav/03a01Fa.wav',
+        ...     version='1.2.0',
+        ...     verbose=False,
+        ... )
+        >>> signal, sampling_rate = read_audio(media[0], end=pd.Timedelta(0.01, unit='s'))
```
frankenjoe, line 134: I completely forgot about this function, it's actually nice to have because I find myself converttng `start` and `end` of an index to `offset` and `duration` to use it with `audiofile.read()`.

frankenjoe:
When I copy the docstring examples, it also copies the trailing `>>>`. Is there a way to change that and only copy the code?

audinterface/core/segment.py
```diff
@@ -131,6 +131,51 @@ class Segment:
     Raises:
         ValueError: if ``resample = True``, but ``sampling_rate = None``
 
+    Example:
+        >>> def segment(signal, sampling_rate, *, win_size=0.2, hop_size=0.1):
+        ...     size = signal.shape[1] / sampling_rate
+        ...     starts = np.arange(0, size - win_size, hop_size)
+        ...     ends = np.arange(win_size, size, hop_size)
+        ...     return pd.MultiIndex.from_tuples(
```
frankenjoe, line 139: Little shorter:

```python
    size = signal.shape[1] / sampling_rate
    starts = pd.to_timedelta(np.arange(0, size - win_size, hop_size), unit='s')
    ends = pd.to_timedelta(np.arange(win_size, size, hop_size), unit='s')
    return pd.MultiIndex.from_tuples(zip(starts, ends), names=['start', 'end'])
```

hagenw:
> When I copy the docstring examples, it also copies the trailing `>>>`. Is there a way to change that and only copy the code?

I forgot that you have to specify this explicitly in the `conf.py` for the plugin. I added it now.

audinterface/core/segment.py
hagenw: Changed accordingly.

PR #49, opened 2022-06-10
Allow single valued context processing function

### LINKED ISSUE #47: Improve audinterface.ProcessWithContext error message for process function with wrong number of return values
I first create an example segmented index (the file has a length of >1s):

```python
import audb
import audformat

db = audb.load('emodb', version='1.2.0', media='wav/03a01Fa.wav', verbose=False)
files = list(db.files) * 3
starts = [0, 0.1, 0.2]
ends = [0.5, 0.6, 0.7]
```

Then I try to run `audinterface.ProcessWithContext` on it.

```python
import audinterface
import numpy as np

def mean(signal, sampling_rate, starts, ends):
    return signal.mean()

interface = audinterface.ProcessWithContext(process_func=mean)
```

But this fails with

```
TypeError                                 Traceback (most recent call last)
<ipython-input-5-9352d0eb840b> in <module>
      6 
      7 interface = audinterface.ProcessWithContext(process_func=mean)
----> 8 interface.process_index(index)

~/git/audeering/audinterface/audinterface/core/process.py in process_index(self, index, root)
    746                 print(file)
    747                 ys[idx] = pd.Series(
--> 748                     self.process_signal_from_index(
    749                         signal, sampling_rate, select,
    750                     ).values,

~/git/audeering/audinterface/audinterface/core/process.py in process_signal_from_index(self, signal, sampling_rate, index)
    789 
    790         # For an empty Series we force the dtype
--> 791         if len(y) == 0:
    792             y = pd.Series(y, index=index, dtype=object)
    793         else:

TypeError: object of type 'numpy.float32' has no len()
```

hagenw:
Closes #47 

This allows processing functions with a single valued return value.

Currently, this is not in line with the docs which state

image

But we seem to allow also that the processing function returns an empty sequence.

An alternative implementation would always check if the number of returned values match the number of segments, but this would break the current behavior for empty sequences.

### DIFF
--- a/audinterface/core/process.py
+++ b/audinterface/core/process.py
@@ -1,3 +1,4 @@
+import collections
 import errno
 import os
 import typing
@@ -819,7 +820,7 @@ class ProcessWithContext:
         y = self(signal, sampling_rate, starts_i, ends_i)
 
         # For an empty Series we force the dtype
-        if len(y) == 0:
+        if isinstance(y, collections.abc.Iterable) and len(y) == 0:
             y = pd.Series(y, index=index, dtype=object)
         else:
             y = pd.Series(y, index=index)
--- a/tests/test_process_with_context.py
+++ b/tests/test_process_with_context.py
@@ -41,6 +41,29 @@ def test_process_func_args():
         )
 
 
+def test_process_func_single_value(tmpdir):
+    # Test context functions that return only a single value
+    # https://github.com/audeering/audinterface/issues/47
+
+    def process_func(signal, sampling_rate, starts, ends):
+        return 1
+
+    process = audinterface.ProcessWithContext(process_func=process_func)
+
+    # create file
+    sampling_rate = 8000
+    signal = np.random.uniform(-1.0, 1.0, (1, 3 * sampling_rate))
+    root = str(tmpdir.mkdir('wav'))
+    file = 'file.wav'
+    path = os.path.join(root, file)
+    af.write(path, signal, sampling_rate)
+    index = audformat.segmented_index([path] * 2, [0, 1], [1, 2])
+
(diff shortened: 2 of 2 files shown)


audinterface/core/process.py
```diff
@@ -819,7 +820,7 @@ def process_signal_from_index(
         y = self(signal, sampling_rate, starts_i, ends_i)
 
         # For an empty Series we force the dtype
-        if len(y) == 0:
+        if isinstance(y, collections.abc.Iterable) and len(y) == 0:
```
frankenjoe, line 823: I guess we added the check for an empty sequence for the special case that the index is empty. But of course check this before we actually do the processing, then we can remove the `if-else` case here.

frankenjoe:
> But we seem to allow also that the processing function returns an empty sequence.

As I said in the comment above I think we added this check for the special case that the `index` is empty. But it was badly implemented of course.

> But we seem to allow also that the processing function returns an empty sequence.

Yes, I think we should do that.

frankenjoe:
The implementation you suggest here makes sense if we see the use case for a model that outputs the same prediction for every segment in a file. But does this really make sense?

hagenw:
> > An alternative implementation would always check if the number of returned values match the number of segments
> 
> Yes, I think we should do that.

I tried to do this, but the it is not so easy as `process_signal_from_index()` is used in both process and process with index. So the easiest way would still be to allow for zero or a single return value as well. But I will further try and submit a new pull request.

hagenw:
I have now implemented an alternative to this pull request in https://github.com/audeering/audinterface/pull/51

frankenjoe:
Closing in favor of #51

PR #51, opened 2022-06-22
ProcessWithContext requires output per segment

### LINKED ISSUE #47: Improve audinterface.ProcessWithContext error message for process function with wrong number of return values

hagenw:
Closes #47 

Alternative implementation to https://github.com/audeering/audinterface/pull/49

This raises an error if the length of the returned sequence of the process function does not match the number of segments for `audinterafce.ProcessWithContext`.

image

image

### DIFF
--- a/audinterface/core/process.py
+++ b/audinterface/core/process.py
@@ -753,7 +753,7 @@ class ProcessWithContext:
         Raises:
             RuntimeError: if sampling rates do not match
             RuntimeError: if channel selection is invalid
-            RuntimeError: if sequence returned by ``self.process_func``
+            RuntimeError: if sequence returned by ``process_func``
                 does not match length of ``index``
 
         .. _audformat: https://audeering.github.io/audformat/data-format.html
@@ -809,7 +809,7 @@ class ProcessWithContext:
         Raises:
             RuntimeError: if sampling rates do not match
             RuntimeError: if channel selection is invalid
-            RuntimeError: if sequence returned by ``self.process_func``
+            RuntimeError: if sequence returned by ``process_func``
                 does not match length of ``index``
             ValueError: if index contains duplicates
 
--- a/audinterface/core/process.py
+++ b/audinterface/core/process.py
@@ -1,3 +1,4 @@
+import collections
 import errno
 import os
 import typing
@@ -806,6 +807,8 @@ class ProcessWithContext:
         Raises:
             RuntimeError: if sampling rates do not match
             RuntimeError: if channel selection is invalid
+            RuntimeError: if sequence returned by ``self.process_func``
+                does not match length of ``index``
             ValueError: if index contains duplicates
 
         .. _audformat: https://audeering.github.io/audformat/data-format.html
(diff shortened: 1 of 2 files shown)


audinterface/core/process.py
```diff
@@ -806,22 +809,30 @@ def process_signal_from_index(
...
-            y = pd.Series(y, index=index, dtype=object)
+        if len(index) == 0:
+            y = pd.Series([], index=index, dtype=object)
         else:
+            starts_i, ends_i = utils.segments_to_indices(
+                signal, sampling_rate, index,
+            )
+            y = self(signal, sampling_rate, starts_i, ends_i)
+            if (
+                    not isinstance(y, collections.abc.Iterable)
```
frankenjoe, line 829: does this capture the empty case or what is it for?

tests/test_process_with_context.py
```diff
@@ -168,6 +184,39 @@ def test_process_index(tmpdir):
...
+            44100,
+            audinterface.utils.signal_index(
+                pd.timedelta_range('0s', '3s', 3),
+                pd.timedelta_range('1s', '4s', 3),
+            ),
+            marks=pytest.mark.xfail(raises=RuntimeError),
+        ),
+        pytest.param(  # process_func returns array with wrong length
+            signal_single_value,
+            lambda sig, fs, starts, ends: [0, 1],
```
frankenjoe, line 211: would it make sense to use a lambda in all cases to avoid introducing functions?

audinterface/core/process.py
hagenw: The first one capture the case of a single returned value, for which we get an error when doing `len(y)`

tests/test_process_with_context.py
hagenw: Switched to using `lambda` in the new tests

audinterface/core/process.py
frankenjoe: According to the type hints the function must always return a sequence:

image

So theoretically we don't have to check for this case :)

hagenw: But we want to raise our custom error message here, this is the reason why we check for it as well.

hagenw: I mean in principle you are right, but as we raise already the custom error, I thought we could also include the case where no sequence is provided.

frankenjoe: Yes, but as long as the user sticks to the type hints, you can safely apply `len(y)`. But ok, doesn't matter.

PR #52, opened 2022-06-22
Add min_signal_dur, max_signal_dur arguments

### LINKED ISSUE #41: Option to set segment limits
I often have the use case that a model expects a minimum signal length and / or I want to set a maximum segment length (e.g. to avoid running out-of-memory). I currently do it like this:

```python
    def process_func(
            self,
            signal: np.ndarray,
            sampling_rate: int,
            min_signal_len: int,
            max_signal_len: int,
    ) -> np.ndarray:

        signal = signal[:, :max_signal_len]
        num_samples = signal.shape[1]
        if num_samples < min_signal_len:
            num_pad = min_signal_len - num_samples
            signal = np.pad(signal, ((0, 0), (0, num_pad)), 'constant')

       # do the actual processing
        
        ...


interface = Process(
    process_func, 
    process_func_args={
        'min_signal_len': 16000,
        'max_signal_len': 160000,
    },
)
```

Since this is a very common use-case, I wonder if it makes sense to add direct support for setting a segment limit that a user can simply do:

```python
    def process_func(
            self,
            signal: np.ndarray,  # satisfies limits
            sampling_rate: int,
    ) -> np.ndarray:

       # do the actual processing
        
        ...


interface = Process(
    process_func, 
    min_signal_len=16000,
    max_signal_len=160000,
)
```

hagenw:
Closes #41 

This adds the arguments `min_signal_dur` and `max_signal_dur` to `audinterface.Process`, `audinterface.Feature`, and `audinterface.Segment`, but not to `audinterafce.ProcessWithContext` as there the process function should take care itself about it.

The signal is cut at the end or zero padded at the end to fullfill the requested requirements.

If both arguments are given it is not checked if `max_signal_dur` > `min_signal_dur`, but `max_signal_dur` is always applied first.

image

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -86,16 +86,24 @@ class Feature:
             Afterwards processing is applied to each segment
         keep_nat: if the end of segment is set to ``NaT`` do not replace
             with file duration in the result
-        min_signal_length: minimum signal length
+        min_signal_dur: minimum signal duration
             required by ``process_func``.
             If value is as a float or integer
             it is treated as seconds.
+            To specify a unit provide as string,
+            e.g. ``'2ms'``.
+            To specify in samples provide as string without unit,
+            e.g. ``'2000'``
             If provided signal is shorter,
             it will be zero padded at the end
-        max_signal_length: maximum signal length
+        max_signal_dur: maximum signal duraton
             required by ``process_func``.
             If value is as a float or integer
             it is treated as seconds.
+            To specify a unit provide as string,
+            e.g. ``'2ms'``.
+            To specify in samples provide as string without unit,
+            e.g. ``'2000'``
             If provided signal is longer,
             it will be cut at the end
         num_workers: number of parallel jobs or 1 for sequential
@@ -155,8 +163,8 @@ class Feature:
             mixdown: bool = False,
             segment: Segment = None,
             keep_nat: bool = False,
-            min_signal_length: Timestamp = None,
-            max_signal_length: Timestamp = None,
+            min_signal_dur: Timestamp = None,
+            max_signal_dur: Timestamp = None,
             num_workers: typing.Optional[int] = 1,
             multiprocessing: bool = False,
             verbose: bool = False,
@@ -227,8 +235,8 @@ class Feature:
             channels=channels,
             mixdown=mixdown,
             segment=segment,
-            min_signal_length=min_signal_length,
-            max_signal_length=max_signal_length,
+            min_signal_dur=min_signal_dur,
+            max_signal_dur=max_signal_dur,
             keep_nat=keep_nat,
             num_workers=num_workers,
             multiprocessing=multiprocessing,
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -86,12 +86,16 @@ class Feature:
             Afterwards processing is applied to each segment
         keep_nat: if the end of segment is set to ``NaT`` do not replace
             with file duration in the result
-        min_signal_length: minimum signal length in samples
+        min_signal_length: minimum signal length
             required by ``process_func``.
+            If value is as a float or integer
+            it is treated as seconds.
             If provided signal is shorter,
             it will be zero padded at the end
-        max_signal_length: maximum signal length in samples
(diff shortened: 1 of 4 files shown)


hagenw:
What I did not do is to provide a `unit` argument, so you can also provide the minimum/maximum in seconds or another unit.
Here, again I was not in favor of adding another argument.

What we could do instead is provide the option to provide them the same way you can provide `start` or `end`:

image

frankenjoe:
> What we could do instead is provide the option to provide them the same way you can provide `start` or `end`:

Yes, I think that would be handy.

frankenjoe:
Supporting only number of samples has the disadvantage that you cannot process signals with different sampling rates.

hagenw:
I switched to use `audinterface.core.typing.Timestamp` as type for `min_signal_length` and `max_signal_length` as we do for `start` and `end`.

audinterface/core/feature.py
```diff
@@ -77,6 +77,18 @@ class Feature:
             Afterwards processing is applied to each segment
         keep_nat: if the end of segment is set to ``NaT`` do not replace
             with file duration in the result
+        min_signal_length: minimum signal length
```
frankenjoe, line 80: Should we call it `min_signal_dur` and `max_signal_dur` for consistency with `win_dur` and `hop_dur`?

audinterface/core/feature.py
```diff
@@ -77,6 +77,18 @@ class Feature:
...
+            required by ``process_func``.
+            If value is as a float or integer
```
frankenjoe, line 82: A string, e.g. '1s'  is also supported maybe we should mention that?

frankenjoe:
Would it make sense to assume number of samples if a string without unit is passed?

hagenw:
> Would it make sense to assume number of samples if a string without unit is passed?

I think we should handle it exactly the same way as for `start` and `end`. At the moment we assume nanoseconds if no unit is provided:

```python
>>> audinterface.core.utils.to_timedelta('1')
Timedelta('0 days 00:00:00.000000001')
```

we could change that, but it would break current behavior for `start` and `end`.

frankenjoe:
> we could change that, but it would break current behavior for start and end.

True but, I doubt anyone has ever provided the value in nanoseconds :) I think we can take the risk :)

If we support samples, we could in principle deprecate the `unit` argument. since it is 'seconds' by default. Until the argument is removed we can capture it's value if explicitly given and convert the argument to it's according string representation.

hagenw:
> > we could change that, but it would break current behavior for start and end.
> 
> True but I doubt anyone has ever provided the value in nanoseconds and I think we can take that risk :)
> 
> If we support samples, we could in principle deprecate the `unit` argument in `Feature`. since it is `'seconds'` by default. Until the argument is removed we can capture it's value if explicitly given and convert the argument to it's according string representation.

At the moment we support `'samples'` as unit only if the `sampling_rate` argument is provided:

```python
        if unit == 'samples' and sampling_rate is None and win_dur is not None:
            raise ValueError(
                "You have specified 'samples' as unit, "
                "but haven't provided a sampling rate."
            )
```
as we use that setting to convert samples to seconds.
An alternative would be to use the sampling rate of the signals we process, but this would require that we add an additional `audiofile.sampling_rate()` call for each file as we need to get start and end in seconds before reading the actual audio.

I would vote for requiring `sampling_rate` as argument whenever we use samples as time input when deprecating `unit`.

frankenjoe:
> I would vote for requiring `sampling_rate` as argument whenever we use samples as time input when deprecating `unit`.

Agree. It anyway does not make much sense to specify a duration in samples if the sampling rate can vary.

audinterface/core/feature.py
hagenw: I switched to `min_signal_dur` and `max_signal_dur`

audinterface/core/feature.py
hagenw: I have now adjusted the docstring to be the same as for `start` or `win_dur`

frankenjoe:
Great, looks good now. Could you please update the description?

hagenw:
> Great, looks good now. Could you please update the description / title?

done

PR #53, opened 2022-06-23
Deprecate unit, specify samples as string

hagenw:
As proposed in https://github.com/audeering/audinterface/pull/52#issuecomment-1164249905 this uses the same time format for `start`, `end`, `win_dur`, `hop_dur` and allows specifying of samples with a string, e.g. `'44100'`.
The `unit` argument of `audinterafce.Feature` was deprecated as it is no longer needed.

image

image

image

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -154,6 +154,22 @@ class Feature:
     ):
         feature_names = audeer.to_list(feature_names)
 
+        # ------
+        # Handle deprecated 'unit' keyword argument
+        def add_unit(dur, unit):
+            if unit == 'samples':
+                return str(dur)
+            else:
+                return f'{dur}{unit}'
+
+        if 'unit' in kwargs:
+            unit = kwargs['unit']
+            if win_dur is not None:
+                win_dur = add_unit(win_dur, unit)
+            if hop_dur is not None:
+                hop_dur = add_unit(hop_dur, unit)
+        # ------
+
         process_func_args = process_func_args or {}
         if kwargs:
             warnings.warn(
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -162,6 +162,8 @@ class Feature:
             message = (
                 "'unit' argument is deprecated "
                 "and will be removed with version '1.2.0'."
+                "The unit can now directly specified "
+                "within the 'win_dur' and 'hop_dur' arguments."
             )
             warnings.warn(
                 message,
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -60,15 +60,24 @@ class Feature:
         sampling_rate: sampling rate in Hz.
             If ``None`` it will call ``process_func`` with the actual
             sampling rate of the signal
-        win_dur: window size in ``unit``,
-            if features are extracted with a sliding window
-        hop_dur: hop size in ``unit``,
+        win_dur: window size,
+            if features are extracted with a sliding window.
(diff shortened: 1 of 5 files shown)


frankenjoe:
Great, I suggest we merge this one first that we can directly apply it to #52

audinterface/core/feature.py
```diff
@@ -117,6 +126,10 @@ class Feature:
         wav/03a01Fa.wav 0 days 0 days 00:00:01.898250 -0.000311  0.082317
 
     """
+    @audeer.deprecated_keyword_argument(
```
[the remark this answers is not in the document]
frankenjoe: Same for `hop_dur` of course.

hagenw:
> Great, I suggest we merge this one first that we can directly apply it to #52

Yes, this was my idea.

audinterface/core/feature.py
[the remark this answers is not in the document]
hagenw: Good point, I added extra code for this.

tests/test_feature.py
```diff
@@ -33,19 +34,42 @@ def features_extractor_sliding_window(signal, _, hop_size):
...
+            audeer.LooseVersion(audinterface.__version__)
+            < audeer.LooseVersion('1.2.0')
+    ):
+        with pytest.warns(UserWarning, match='is deprecated'):
+            interface = audinterface.Feature(
+                ['a'],
+                win_dur=1000,
+                unit='samples',
+            )
+            assert interface.win_dur == '1000'
```
hagenw, line 48: This test fails as `kwargs` inside `audinterface.Feature` is `{}` and does not contain `unit`. Any idea why this could be the case?

frankenjoe: It's removed by the decorator. Maybe the `audeer.deprecated_keyword_argument` does not forward it?

frankenjoe: Yes, it's removed here: https://github.com/audeering/audeer/blob/9798923c7ef1f8dfdadce6e21c65678688fcd6aa/audeer/core/utils.py#L166

hagenw: So the question is if we should modify the behavior of `@audeer.deprecated_keyword_argument` and remove it only from `kwargs` when a new keyword arg is specified, or if I should simply use custom code here instead.

frankenjoe: If within the mapping function we had access to the other arguments, we could directly modify `win_dur` there. This would probably be the cleanest solution. But I think the workaround is fine for the moment.

hagenw: I created https://github.com/audeering/audeer/issues/69

frankenjoe: > is not a solution as it will suggest to the user she/he should use `unit` instead of `unit`.

Hehe, good that you noticed that :)

audinterface/core/utils.py
```diff
@@ -289,8 +290,43 @@ def to_array(value: typing.Any) -> np.ndarray:
...
+        if isinstance(time, str):
+            # ensure we have a str and not numpy.str_
+            time = str(time)
+            # string without unit represents samples
+            if all(t.isdigit() for t in time):
+                if sampling_rate is None:
+                    raise ValueError(
+                        "You have to provide 'sampling_rate' "
+                        "when specifying the duration in samples "
+                        "as you did with '{time}'"
```
frankenjoe, line 315:
```suggestion
                        f"as you did with '{time}'."
```

Issue #54, opened 2022-06-23
Feature: add option to call process function on a sliding window


frankenjoe:
Currently, when working with a sliding window, the user is responsible to handle the windowing. Unless in the case that a function from an external library is called that already implements a sliding window, this makes it quite inconvenient for a user to work with this feature. E.g. the following (not very efficient) code is needed to just calculate the mean over a sliding window:

```python
def mean(signal, sampling_rate, win_dur, hop_dur):
    win_dur = int(sampling_rate * win_dur)
    hop_dur = int(sampling_rate * hop_dur)
    steps = (signal.shape[-1] - win_dur) // hop_dur
    pos = 0
    y = []
    for _ in range(steps):
        win = signal[0, pos:pos+win_dur]
        y.append(win.mean())
        pos += hop_dur
    return np.atleast_2d(y)

win_dur=1.0
hop_dur=0.5
interface = audinterface.Feature(
    'mean',
    process_func=mean,
    process_func_args={
        'win_dur': win_dur,
        'hop_dur': hop_dur,
    },
    win_dur=win_dur,
    hop_dur=hop_dur,
)
```

And it's also easy to mess up, e.g. the following would work, but produce wrong timestamps:

```python
win_dur=1.0
hop_dur=0.5
interface = audinterface.Feature(
    'mean',
    process_func=mean,
    process_func_args={
        'win_dur': win_dur,
        'hop_dur': hop_dur,
    },
    win_dur=win_dur * 2,
    hop_dur=hop_dur * 2,
)
```

I think it would be very convenient to offer an option that the windowing is handled internally so that the user only has to provide the actual processing code, e.g.:

```
def mean(signal, sampling_rate):
    return signal.mean()

win_dur=1.0
hop_dur=0.5
interface = audinterface.Feature(
    'mean',
    process_func=mean,
    process_func_sliding=True,  # apply sliding window on signal and forward only single frames to processing function
    win_dur=win_dur,
    hop_dur=hop_dur,
)
```

hagenw:
Yes, seems to make sense to provide this.

`process_func_sliding=True` seems to be not the optimal name as it would suggest to me that it indicates that the process function applies already a sliding window, which is not the case ;)

frankenjoe:
> `process_func_sliding=True` seems to be not the optimal name as it would suggest to me that it indicates that the process function applies already a sliding window, which is not the case ;)

Yes, I'm also not happy with it. Any suggestions?

[...]

PR #55, opened 2022-06-24
Remove deprecated kwargs for process_func_args

hagenw:
When introducing `process_func_args` in version 0.8.0 beginning of the year we stated that it will be removed in version 1.0.0.

As I think we should release the new version as 1.0.0, this removes it.

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -328,16 +328,6 @@ class Feature:
                     dtype=object,
                 )
 
-        process_func_args = process_func_args or {}
-        if kwargs:
-            warnings.warn(
-                utils.kwargs_deprecation_warning,
-                category=UserWarning,
-                stacklevel=2,
-            )
-            for key, value in kwargs.items():
-                process_func_args[key] = value
-
         if win_dur is None and hop_dur is not None:
             raise ValueError(
                 "You have to specify 'win_dur' if 'hop_dur' is given."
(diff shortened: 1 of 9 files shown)


hagenw:
Note, that in `audinterface.Feature` I cannot yet remove `**kwargs` from `__init__()` as we need it for support of the deprecated `unit` keyword argument.

frankenjoe:
Are you sure we should immediately target `1.0.0`? Maybe we should release `0.9.0` in between. Also it would be nice to have #54 before we release `1.0.0`.

hagenw:
I guess, you are right. We added a lot of new stuff, so maybe releasing 0.9.0 first makes sense. On the other hand, it is very unlikely that we will break the API anyway.

But I will set this pull request to WIP for now.

hagenw:
@frankenjoe should we target 1.0.0 as next release or go with 0.10.0?

frankenjoe:
Maybe we should wait with `1.0.0` until we know what new functionality we will need to support storing features in `audformat`. At least reading the features file should be supported, but maybe also their creation. Btw: we have not created an issue regarding feature support in `audformat` yet, will do now and link it here.

frankenjoe:
https://github.com/audeering/audformat/issues/321

hagenw:
Yes, please go ahead with 0.10.0

hagenw:
I removed `WIP` as I think we can target version 1.0.0 with the next release?

frankenjoe:
There's one more issue I'd like to tackle before, see https://github.com/audeering/audinterface/issues/100. I have already some code that I was using to avoid this problem I can build on. So it will hopefully not take too long. I'll try to prepare a PR soonish.

hagenw:
Cool, don't worry we don't need to rush with this.

hagenw:
I agree, I have rebased `main` and you can go ahead here.

PR #56, opened 2022-06-24
Clean up init() functions

frankenjoe:
* Changes the order of some keyword arguments
* Orders attributes alphabetically
* Moves `ProcessWithContext` to own file
* Adds missing test for `ProcessWithContext` with multiple channels


hagenw:
Thanks for cleaning up.

PR #57, opened 2022-06-27
Add utils.window()

frankenjoe:
First step towards #54 

image

### DIFF
--- a/audinterface/core/utils.py
+++ b/audinterface/core/utils.py
@@ -346,8 +346,8 @@ def window(
 
     Args:
         signal: input signal in shape
-            ``(samples, )``
-            or``(channels, samples)``
+            ``(samples,)``
+            or ``(channels, samples)``
         sampling_rate: sampling rate in Hz
         win_dur: window duration,
             if value is as a float or integer
--- a/audinterface/core/utils.py
+++ b/audinterface/core/utils.py
@@ -10,7 +10,10 @@ import audformat
 import audresample
 import audiofile as af
 
-from audinterface.core.typing import Timestamps
+from audinterface.core.typing import (
+    Timestamp,
+    Timestamps,
+)
 
 
 def assert_index(obj: pd.Index):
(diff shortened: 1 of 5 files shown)


audinterface/core/utils.py
```diff
@@ -331,3 +334,105 @@ def convert_samples_to_seconds(time):
...
+
+    shape = (signal.shape[0], signal.shape[1] - win_length + 1, win_length)
+    strides = (signal.strides[0], signal.strides[1], signal.strides[1])
+    signal = np.lib.stride_tricks.as_strided(
+        signal,
+        strides=strides,
+        shape=shape,
+    )[:, 0::hop_length]
+
+    signal = signal.swapaxes(0, 1)  # make frames first axis
```
frankenjoe, line 436: Otherwise we would return `(channels, frames, samples)`, but I think `(frames, channels, samples)` makes more sense cause now the user can simply do `frames[idx]` to select a specific frames.

hagenw: In `audinterface.Feature` we request that a process function with a sliding window returns `(channels, features, frames)`, so I guess here `(channels, samples, frames)` would make the most sense?

hagenw: But I also see the point of `for frame in frames`. Maybe you could provide an example how you would use `window` to create an example process function creating the mean per frame.

frankenjoe: Right, there we have `(channels, features, frames)`. So maybe you are right and it makes more sense to switch the last two axis then.

tests/test_utils.py
```diff
@@ -126,3 +127,187 @@ def test_create_segmented_index(starts, ends):
             assert index.get_level_values(
                 audformat.define.IndexField.END
             ).tolist() == [pd.NaT] * len(starts)
+
+
+@pytest.mark.parametrize(
+    'signal, sampling_rate, win_dur, hop_dur, expected',
```
hagenw, line 133: You never specified a `sampling_rate` other than 1, so maybe remove it from the parameters?

tests/test_utils.py
```diff
@@ -126,3 +127,187 @@ def test_create_segmented_index(starts, ends):
...
+                    [30, 31, 32, 33],
+                ],
+            ),
+            1, 1, 1,
+            np.array(
+                [
+                    [[10], [20], [30]],
+                    [[11], [21], [31]],
+                    [[12], [22], [32]],
+                    [[13], [23], [33]]],
```
hagenw, line 197:
```suggestion
                    [[13], [23], [33]],
                ],
```

audinterface/core/utils.py
```diff
@@ -331,3 +334,105 @@ def convert_samples_to_seconds(time):
         return pd.to_timedelta(times, unit='s')
     except ValueError:  # catches values like '1s'
         return pd.to_timedelta(times)
+
+
+def window(
```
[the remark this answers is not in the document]
frankenjoe: Yes, thought about that, too. So maybe it's indeed better to be more explicit here.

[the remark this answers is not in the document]
frankenjoe: done

audinterface/core/utils.py
```diff
@@ -331,3 +334,105 @@ def convert_samples_to_seconds(time):
...
+        signal: np.ndarray,
+        sampling_rate: int,
+        win_dur: Timestamp,
+        hop_dur: Timestamp,
+):
+    r"""Reshape signal by applying a sliding window.
```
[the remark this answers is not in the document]
frankenjoe: Done (see description)

audinterface/core/utils.py
hagenw: Thanks, so from that example `(frames, channels, samples)` makes sense. On the other hand if we switch to `(channels, features, frames)` we could directly use `sliding_window()` as a processing function.

tests/test_utils.py
frankenjoe: Yes we can do it, but let's do it as a last step once we are happy with the tests and don't see the need to test other sampling rates.

audinterface/core/utils.py
hagenw: The following should also work, but it is also not very obvious:
```python
y = frames.mean(axis=1).mean(axis=0)
```

But I agree with you that it is better to stick with our convention and return it as `(channels, features, frames)`.

frankenjoe: Actually, if we return `(channels, features, frames)` the following works:

```
def mean(signal, sampling_rate, win_dur, hop_dur):
        signal,
        sampling_rate,
        win_dur,
        hop_dur,
    )
    return frames.mean(axis=1)
```

frankenjoe: And with `frames.mean(axis=1, keepdims=True)` it even works for multiple channels, e.g.:

```python
...
    'mean',
    process_func=mean,
    process_func_args={
        'win_dur': win_dur,
        'hop_dur': hop_dur,
    },
    win_dur=win_dur,
    hop_dur=hop_dur,
    channels=[0, 0],
)
```
```
                                                                 mean-0    mean-1
file            start                  end                                       
wav/03a01Fa.wav 0 days 00:00:00        0 days 00:00:01        -0.000329 -0.000329
                0 days 00:00:00.500000 0 days 00:00:01.500000 -0.000285 -0.000285
wav/03a01Nc.wav 0 days 00:00:00        0 days 00:00:01         0.000039  0.000039
                0 days 00:00:00.500000 0 days 00:00:01.500000 -0.000412 -0.000412
wav/03a01Wa.wav 0 days 00:00:00        0 days 00:00:01        -0.000284 -0.000284
                0 days 00:00:00.500000 0 days 00:00:01.500000 -0.000318 -0.000318
```

frankenjoe: I now changed to `(channels, features, frames)` (see description).

tests/test_utils.py
hagenw: We still have only `1` as value. But I guess it will make it easier to extend the test at a later stage if we stick with `sampling_rate` as parameter here.

PR #58, opened 2022-06-28
Process with sliding window

frankenjoe:
Relates to #54 

Adds option to use `Process` with a sliding window.

### API

image

...

image

### Usage

image

### DIFF
--- a/audinterface/core/process.py
+++ b/audinterface/core/process.py
@@ -36,6 +36,24 @@ class Process:
         resample: if ``True`` enforces given sampling rate by resampling
         channels: channel selection, see :func:`audresample.remix`
         mixdown: apply mono mix-down on selection
+        win_dur: window duration,
+            if features are extracted with a sliding window.
+            If value is as a float or integer
+            it is treated as seconds.
+            To specify a unit provide as string,
+            e.g. ``'2ms'``.
+            To specify in samples provide as string without unit,
+            e.g. ``'2000'``
+        hop_dur: hop duration,
+            if features are extracted with a sliding window.
+            This defines the shift between two windows.
+            If value is as a float or integer
+            it is treated as seconds.
+            To specify a unit provide as string,
+            e.g. ``'2ms'``.
+            To specify in samples provide a string without unit,
+            e.g. ``'2000'``.
+            Defaults to ``win_dur / 2``
         min_signal_dur: minimum signal length
             required by ``process_func``.
             If value is as a float or integer
@@ -96,8 +114,18 @@ class Process:
         file             start   end
         wav/03a01Fa.wav  0 days  0 days 00:00:01.898250    -0.000311
         dtype: float32
+        >>> interface = Process(
+        ...     process_func=mean,
+        ...     win_dur=1.0,
+        ...     hop_dur=0.5,
+        ... )
+        >>> interface.process_index(index, root=db.root)
+        file             start                   end
+        wav/03a01Fa.wav  0 days 00:00:00         0 days 00:00:01          -0.000329
+                         0 days 00:00:00.500000  0 days 00:00:01.500000   -0.000285
+        dtype: float32
 
-    """
+    """  # noqa: E501
     def __init__(
             self,
             *,
@@ -108,6 +136,8 @@ class Process:
             resample: bool = False,
             channels: typing.Union[int, typing.Sequence[int]] = None,
             mixdown: bool = False,
+            win_dur: Timestamp = None,
+            hop_dur: Timestamp = None,
             min_signal_dur: Timestamp = None,
             max_signal_dur: Timestamp = None,
             segment: Segment = None,
@@ -139,11 +169,21 @@ class Process:
                 'sampling_rate has to be provided for resample = True.'
             )
 
+        if win_dur is None and hop_dur is not None:
+            raise ValueError(
+                "You have to specify 'win_dur' if 'hop_dur' is given."
+            )
+        if win_dur is not None and hop_dur is None:
+            hop_dur = utils.to_timedelta(win_dur, sampling_rate) / 2
+
         self.channels = channels
         r"""Channel selection."""
 
         self.keep_nat = keep_nat
         r"""Keep NaT in results."""
+
+        self.hop_dur = hop_dur
+        r"""Hop duration."""
 
         self.max_signal_dur = max_signal_dur
         r"""Maximum signal length."""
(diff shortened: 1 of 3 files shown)


frankenjoe:
I removed the WIP now, ready for inspection :)

Once we have this in place, it should be straight-forward to solve #54 - either the processing function takes care of the sliding window itself then we have the current behavior - of if the user disables it - we simply forward `win_dur` and `hop_dur` to `Process`.

hagenw:
Thanks for the effort, I will review next week.

audinterface/core/process.py
```diff
@@ -36,6 +36,24 @@ class Process:
...
         channels: channel selection, see :func:`audresample.remix`
         mixdown: apply mono mix-down on selection
+        win_dur: window duration,
+            if processing should be applied on a sliding window.
+            If value is as a float or integer
+            it is treated as seconds.
+            To specify a unit provide as string,
+            e.g. ``'2ms'``.
+            To specify in samples provide as string without unit,
+            e.g. ``'2000'``
```
hagenw, line 46: I'm always not sure what the best way is to handle the following problem:
We now repeat how you can specify duration four times in the docstring (`win_dur`, `hop_dur`, `min_signal_dur`, `max_signal_dur`), and the same in other docstrings as well. So it would make sense to write it down once, maybe at the top of each docstring, or as a subsection in the user documentation.
On the other hand, if I look for a particular argument of the object, I will directly look for it and not read the top of the docstring. So from a user perspective the current "implemenation" of the docstring makes sense, but it is not the best solution for maintaining.

Do you have a preference (independent of that I would suggest to not handle a possible solution in this pull request, but go with your current solution)?

audinterface/core/process.py
```diff
@@ -698,16 +772,33 @@ def __call__(
             self.channels,
             self.mixdown,
         )
-        if self.process_func_is_mono:
-            return [
-                self.process_func(
-                    np.atleast_2d(channel),
+
+        def _helper(x):
```
hagenw, line 776: Maybe call it `_process` instead of `_helper`, but on the other hand we have already a lot of functions/methods called process ;)

docs/usage.rst
```diff
@@ -95,6 +95,19 @@ as a :class:`pandas.Series`.
...
+
+.. jupyter-execute::
+
+    interface = audinterface.Process(
+        process_func=rms,
+        win_dur=1.0,
+        hop_dur=0.5,
+    )
+    y = interface.process_files(files)
+    y
```
hagenw, line 110: Nice, now this becomes very easy.

tests/test_process.py
```diff
@@ -1106,6 +1106,142 @@ def test_process_signal_min_max(
...
+    )
+
+    # process signal
+    y = process.process_signal(
+        signal,
+        sampling_rate,
+        start=start,
+        end=end,
+    )
+    pd.testing.assert_series_equal(y, expected, atol=1.0e-4)
```
hagenw, line 1209: Interesting, this first of the five tests works for me without `atol`. I'm also a little bit surprised that `numpy` is not able to end up with our easy numbers, but I guess this is how it is ;)

Should we maybe define the tolerance in a single variable in the parametrisation, where you specify `'process_func, signal, sampling_rate'`?

tests/test_process.py
```diff
@@ -1106,6 +1106,142 @@ def test_process_signal_min_max(
...
+        (
+            1.0, 5.0, 4, 2,
+            pd.Series(
+                [0.25],
+                audinterface.utils.signal_index(1, 5),
+                dtype=np.float32,
+            ),
+        ),
+        (
+            1.0, 2.0, 4, 2,
```
hagenw, line 1164: Could you also add a test for the case where the window is long enough, but `start` is out of bound, e.g. `9.0, 15.0, 4, 2`?

tests/test_process.py
```diff
@@ -1106,6 +1106,142 @@ def test_process_signal_min_max(
...
+                [0, 0.5, 1],
+                audinterface.utils.signal_index(
+                    [0, 2, 4],
+                    [4, 6, 8],
+                ),
+                dtype=np.float32,
+            ),
+        ),
+        (
+            None, None, 4, 2,
```
hagenw, line 1134: Would be nice to also have one test where `hop_dur` is not `win_dur / 2`.

hagenw (commented):
Looks good, I just have a few minor comments.

audinterface/core/process.py
frankenjoe: Would it make sense to add `utils.to_timedelta()` to the API and then simply link to it in those places?

tests/test_process.py
frankenjoe: Done, I added a couple of more tests like `hop_dur == win_dur` and `hop_dur > win_dur`.

tests/test_process.py
frankenjoe: Done.

tests/test_process.py
frankenjoe: The problem comes from rounding errors when storing the input signal to a WAV file. I now changed to 32-bit depth when storing the file, and now it works without `atol`.

audinterface/core/process.py
hagenw: Created https://github.com/audeering/audinterface/issues/59

audinterface/core/process.py
hagenw: OK, then let's stick with your current solution.

tests/test_process.py
hagenw: Ah, great.

tests/test_process.py
hagenw: Thanks, I was also curious to see if it fails as `start` and `end` are out of bound or if it returns an empty array. Seems to be the later. Is this behavior the same, when not using a sliding window, or do we then get an error for out of bound `start` and `end` values?

tests/test_process.py
hagenw: I cannot find `hop_dur > win_dur`, but you now have two times the test with `start=None, end=None, win_dur=4, hop_dur=2`

tests/test_process.py
frankenjoe: > when not using a sliding window, or do we then get an error for out of bound start and end values

In that case an empty array is passed to the processing function so it depends how it is handled by the function.

tests/test_process.py
frankenjoe: Thanks for spotting. I fixed it.

tests/test_process.py
hagenw: Makes sense

tests/test_process.py
hagenw: Cool, now everything looks fine.

PR #60, opened 2022-07-05
Feature: add process_func_applies_sliding_window argument

### LINKED ISSUE #54: Feature: add option to call process function on a sliding window

frankenjoe:
Closes #54 

### API

image

### Tutorial

image

image

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -32,8 +32,8 @@ class Feature:
     If ``win_dur`` is specified ``start`` and ``end`` indices
     are referred from the original ``start`` and ``end`` arguments
     and the window positions.
-    Otherwise, the original ``start`` and ``end`` indices
-    are kept.
+    If ``win_dur`` is ``None``,
+    the original ``start`` and ``end`` indices are kept.
     If
     ``process_func_applies_sliding_window``
     is set to ``True``
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -57,6 +57,18 @@ class Feature:
             function
         process_func_is_mono: apply ``process_func`` to every channel
             individually
+        process_func_applies_sliding_window:
+            if ``True``
+            the processing function receives
+            whole segments and is responsible
+            for applying a sliding window itself.
+            If ``False``,
+            the sliding window is applied internally
+            and the processing function
+            receives invidual frames instead.
+            Applies only if
+            features are extracted in a framewise manner
+            (see ``win_dur`` and ``hop_dur``)
         sampling_rate: sampling rate in Hz.
             If ``None`` it will call ``process_func`` with the actual
             sampling rate of the signal
@@ -155,6 +167,7 @@ class Feature:
             process_func: typing.Callable[..., typing.Any] = None,
             process_func_args: typing.Dict[str, typing.Any] = None,
             process_func_is_mono: bool = False,
+            process_func_applies_sliding_window: bool = True,
             sampling_rate: int = None,
             resample: bool = False,
             channels: typing.Union[int, typing.Sequence[int]] = 0,
@@ -244,6 +257,8 @@ class Feature:
             resample=resample,
             channels=channels,
             mixdown=mixdown,
+            win_dur=None if process_func_applies_sliding_window else win_dur,
+            hop_dur=None if process_func_applies_sliding_window else hop_dur,
             min_signal_dur=min_signal_dur,
             max_signal_dur=max_signal_dur,
             segment=segment,
@@ -276,6 +291,10 @@ class Feature:
 
         self.process = process
         r"""Processing object."""
+
+        self.process_func_applies_sliding_window = \
+            process_func_applies_sliding_window
+        r"""Controls if processing function applies sliding window."""
 
         self.verbose = verbose
         r"""Show debug messages."""
@@ -563,7 +582,7 @@ class Feature:
 
         if self.process.process_func_is_mono:
             # when mono processing is turned on
-            # the channel dimension has to be 1
+            # the channel dimension has to be 1,
             # so we would usually omit it,
             # but since older versions required
             # a channel dimension we have to
@@ -579,6 +598,11 @@ class Feature:
                 # (channels, 1, features)
                 # -> (channels, features)
                 features = features.squeeze(axis=1)
+
+            if self.win_dur and not self.process_func_applies_sliding_window:
+                # (channels, features)
+                # -> (channels, features, 1)
+                features = features.reshape(self.num_channels, -1, 1)
 
         if features.ndim > 3:
             raise RuntimeError(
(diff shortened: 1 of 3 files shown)


tests/test_feature.py
```diff
@@ -910,15 +946,38 @@ def test_signal_sliding_window(win_dur, hop_dur):
...
+
+    interface = audinterface.Feature(
+        feature_names='mean',
+        process_func=mean_sliding_window,
+        process_func_args={
+            'win_dur': 0.5,
+            'hop_dur': 0.25,
+        },
+        process_func_is_mono=False,
+        process_func_applies_sliding_window=True,
```
hagenw, line 967: Why do we need the `process_func_applies_sliding_window` argument?
We have set the window and hop duration anyway as parameter of the sliding window. Or is it just to add a check to ignore the `win_dur` and `hop_dur` arguments inside `Feature`?

frankenjoe: We need some way to tell `Feature` whether the process function will apply the sliding window or if this should be handled internally. So do you propose we should derive that from the arguments to the processing function?

frankenjoe: The other problem is that even if the processing function does not apply the sliding window itself, it may still need the information about the window and hop duration. In that case it would be a bad idea to automatically derive the strategy from the fact that the processing function has a `win_dur` and `hop_dur` argument. So guess it's safer to introduce an argument to control the behavior.

However, I now implemented that the `win_dur` and `hop_dur` arguments are automatically added to `process_func_args` if the processing function expects such arguments and they are not explicitly set by the user. Now:

```python
db = audb.load(
    'emodb',
    version='1.2.0',
    format='wav',
    mixdown=True,
    sampling_rate=16000,
    full_path=False,
)


def mean(signal, sampling_rate, win_dur, hop_dur):
        signal,
        sampling_rate,
        win_dur,
        hop_dur,
    )
    return frames.mean(axis=1, keepdims=True)


win_dur=1.0
hop_dur=0.5

    'mean',
    process_func=mean,
    process_func_applies_sliding_window=True,
    process_func_args={
        'win_dur': win_dur,
        'hop_dur': hop_dur,
    },
    win_dur=win_dur,
    hop_dur=hop_dur,
)
interface.process_files(db.files[:3], root=db.root)
```
```
                                                                   mean
file            start                  end                             
wav/03a01Fa.wav 0 days 00:00:00        0 days 00:00:01        -0.000329
                0 days 00:00:00.500000 0 days 00:00:01.500000 -0.000285
wav/03a01Nc.wav 0 days 00:00:00        0 days 00:00:01         0.000039
                0 days 00:00:00.500000 0 days 00:00:01.500000 -0.000412
wav/03a01Wa.wav 0 days 00:00:00        0 days 00:00:01        -0.000284
                0 days 00:00:00.500000 0 days 00:00:01.500000 -0.000318
```

can be simplified to:

```python
...
    'mean',
    process_func=mean,
    process_func_applies_sliding_window=True,
    win_dur=win_dur,
    hop_dur=hop_dur,
)
...
```

At least there is a way to avoid setting `win_dur` and `hop_dur` twice.

I would then recommend that we immediately deprecate the default value of `process_func_applies_sliding_window` from `True` to `False`.

tests/test_feature.py
```diff
@@ -847,78 +870,93 @@ def test_process_index(tmpdir):
...
-            marks=pytest.mark.xfail(raises=RuntimeError),
-        ),
+        (0.5, 0.25),
+        (0.5, 0.5),
+        (0.25, 0.5),
+        (pd.to_timedelta(1, unit='s'), pd.to_timedelta(0.5, unit='s')),
+        ('4000', '2000'),
+        ('500ms', '250ms'),
+        ('500milliseconds', '250milliseconds'),
+        (f'{SAMPLING_RATE // 2}', f'{SAMPLING_RATE // 4}'),
```
hagenw, line 892: In the tests before we also test `hop_dur=None`, did you omit it on purpose now?

tests/test_feature.py
hagenw: > What's also unfortunate is that - again for backward compatibility - we have to assume by default that the processing function will apply the sliding window, so the default behavior is different from what we have in Process.

In `Process` we don't have the `process_func_applies_sliding_window`, so I guess we don't need to have this argument there and the process func can apply a sliding window or not if we set `win_dur=None`?

frankenjoe: In `Process` such an argument would not make sense, since we do assume that the returned values have a certain shape, i.e. we cannot know which axis should be expand into frames.

tests/test_feature.py
frankenjoe: Oh right, I re-added the test.

audinterface/core/feature.py
```diff
@@ -21,18 +22,35 @@
...
+    the returned object has ``num_channels * num_features`` columns.
     It will have one row per file or signal.
+
     If features are extracted using a sliding window,
     each window will be stored as one row.
     If ``win_dur`` is specified ``start`` and ``end`` indices
     are referred from the original ``start`` and ``end`` arguments
     and the window positions.
     Otherwise, the original ``start`` and ``end`` indices
     are kept.
```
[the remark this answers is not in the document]
frankenjoe: Changed to:

image

PR #62, opened 2022-07-08
Add audinterface.utils.to_timedelta()

### LINKED ISSUE #59: Add to_timedelta() to the API
As discussed in https://github.com/audeering/audinterface/pull/58#discussion_r912940481 it would be handy to have `utils.to_timedelta()` as part of the public API. This would allow:
* reuse it in other packages
* have a single point for documenting its usage to which we can link from other time arguments like `win_dur`

hagenw:
Closes #59 

~~I also adjusted the return type from `typing.Sequence[pd.Timedelta]` to `pd.TimedeltaIndex`.
I'm also fine with changing the output of the function to the return type we mentioned before, but I'm not sure if this will break the behavior inside `audinterface`.~~

image

---

I updated related docstrings to link now to this new utility function:

image

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -94,26 +94,26 @@ class Feature:
         channels: channel selection, see :func:`audresample.remix`
         win_dur: window duration,
             if features are extracted with a sliding window.
-            If value is as a float or integer
+            If value is a float or integer
             it is treated as seconds.
             See :func:`audinterface.utils.to_timedelta` for further options
         hop_dur: hop duration,
             if features are extracted with a sliding window.
             This defines the shift between two windows.
-            If value is as a float or integer
+            If value is a float or integer
             it is treated as seconds.
             See :func:`audinterface.utils.to_timedelta` for further options.
             Defaults to ``win_dur / 2``
         min_signal_dur: minimum signal duration
             required by ``process_func``.
-            If value is as a float or integer
+            If value is a float or integer
             it is treated as seconds.
             See :func:`audinterface.utils.to_timedelta` for further options.
             If provided signal is shorter,
             it will be zero padded at the end
         max_signal_dur: maximum signal duraton
             required by ``process_func``.
-            If value is as a float or integer
+            If value is a float or integer
             it is treated as seconds.
             See :func:`audinterface.utils.to_timedelta` for further options.
             If provided signal is longer,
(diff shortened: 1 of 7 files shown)


audinterface/core/utils.py
```diff
@@ -401,16 +395,47 @@ def to_array(value: typing.Any) -> np.ndarray:
 
 
 def to_timedelta(
-        times: Timestamps,
+        durations: Timestamps,
         sampling_rate: int = None,
-) -> typing.Union[pd.Timedelta, typing.Sequence[pd.Timedelta]]:
-    r"""Convert time value to pd.Timedelta.
+) -> typing.Union[pd.Timedelta, typing.List[pd.Timedelta]]:
+    r"""Convert time value to :class:`pandas.Timedelta`.
```
frankenjoe, line 401:
```suggestion
    r"""Convert time value(s) to :class:`pandas.Timedelta`.
```

frankenjoe: Or maybe better "duration value(s)" as we call it in the arguments.

audinterface/core/utils.py
```diff
@@ -401,16 +395,47 @@ def to_array(value: typing.Any) -> np.ndarray:
...
 
     If time is given as string without unit,
```
frankenjoe, line 403:
```suggestion
    If duration is given as string without unit,
```

audinterface/core/utils.py
```diff
@@ -401,16 +395,47 @@ def to_array(value: typing.Any) -> np.ndarray:
...
     it is treated as samples
     and requires that ``'sampling_rate'`` is not ``None``.
 
-    """
+    Args:
+        durations: duration value(s).
+            If duration value is as a float or integer
```
frankenjoe, line 409:
```suggestion
            If value is a float or integer
```

audinterface/core/utils.py
```diff
@@ -401,16 +395,47 @@ def to_array(value: typing.Any) -> np.ndarray:
...
+            To specify a unit provide as string,
+            e.g. ``'2ms'``.
+            To specify in samples provide as string without unit,
+            e.g. ``'2000'``
+        sampling_rate: sampling rate in Hz.
+            Needs to be provided
+            if any duration value is provided in samples
+
+    Returns:
+        duration values as :class:`pandas.Timedelta` objects
```
frankenjoe, line 420:
```suggestion
        duration value(s) as :class:`pandas.Timedelta` objects
```

audinterface/core/feature.py
```diff
@@ -96,38 +96,26 @@ class Feature:
             if features are extracted with a sliding window.
             If value is as a float or integer
```
frankenjoe, line 97:
```suggestion
            If value is a float or integer
```

Please also change in all the other places.

hagenw: Thanks for spotting, I fixed it at all places.

frankenjoe: Thanks, can you please update the description.

hagenw: done

PR #63, opened 2022-07-08
Deprecate default value of process_func_applies_sliding_window with 1.2.0

### LINKED ISSUE #61: Change default value of `process_func_applies_sliding_window` to `False`
See https://github.com/audeering/audinterface/pull/60#discussion_r916563773

frankenjoe:
Closes #61


hagenw:
If I see it correctly you now have to provide `process_func_applies_sliding_window` explicitly every time in order to avoid the deprecation warning, which seems to me very annoying. Maybe we should only show it when `win_dur` is not `None`?

frankenjoe:
Good idea, I changed it accordingly and added an example to the description.

PR #66, opened 2022-07-12
DOC: better HTML representations of tables

hagenw:
This ensures that the HTML output of `df`, `y`, `index` in the usage section are independent of the root path of the underlying data.
In addition, the index presentation is fixed to look the same as for dataframe or series.

Example:

image

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -3,7 +3,6 @@
     :hide-code:
     :hide-output:
 
-    import os
     import pandas as pd
 
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -3,23 +3,41 @@
     :hide-code:
     :hide-output:
 
+    import os
     import pandas as pd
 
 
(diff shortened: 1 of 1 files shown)


frankenjoe:
Alternatively we can do:

```
db = audb.load(..., full_path=False)
```

and then pass `root=audb.root` when we process files, index, etc.

hagenw:
> Alternatively we can do:
> 
> ```
> db = audb.load(..., full_path=False)
> ```
> 
> and then pass `root=db.root` when we process files, index, etc.

I'm not in favor of this as users tend to copy and paste examples and then they think they always have to provide `root` whereas it would be much easier without.

frankenjoe:
> I'm not in favor of this as users tend to copy and paste examples and then they think they always have to provide `root` whereas it would be much easier without.

I often use `root` as this way the result is portable. But I'm also fine with doing it replacing the manually.

Issue #25, opened 2021-07-30
Process.process_table


ATriantafyllopoulos:
@hagenw @frankenjoe there is sometimes the need for a model to utilise the information in a table, instead of simply accessing the audio files.

For example, for NLP fusion, it makes sense to run the transcription once (as it takes ages), store it in a table, and then run several fusion models on the table containing the transcription. I guess similar use-cases would come for speaker-dependent or gender-dependent model or whatever.

I plan to add this functionality in the fusion package I am creating now, but do you think it makes sense to have it here too?

P.S. Maybe it could also be used with feature tables as well.

frankenjoe:
A possible solution is implemented in #67

hagenw:
This is now implemented with #67 by providing the running index `idx` as a special argument.
You just need to add it to your process function and you can access it.
For the issue of getting transcriptions from a dataframe inside the processing function, you can now do the following (after we release a new version of `audinterface`):

```python
def process_func(signal, sampling_rate, idx, df):
    transcription = df.iloc[0]['transcription']
    ...

interface = audinterface.Process(
    ...,
    process_func_args={'df': df},
)
```
```

PR #67, opened 2022-07-12
Support special arguments to process function

### LINKED ISSUE #65: Option to pass index to process function
In some situations it can be useful to pass an index in addition to signal and sampling rate to the processing function. Similar to the `with_indices` argument in the [Dataset.map()](
https://huggingface.co/docs/datasets/v1.1.1/package_reference/main_classes.html#datasets.Dataset.map) method from Huggingface:

image

Possible use-case:

```python
model = audonnx.load(...)
db = audb.load(...)
files = db['files'].files
gender = db['files']['gender'].get()

def process_func(signal, sampling_rate, idx):
   y = model(signal, sampling_rate)
   if gender[idx] == 'female':
       return y[0]  # return output trained on female samples
   else:
       return y[1]  # return output trained on male samples

interface = audinterface.Process(..., with_indices=True)
interface.process_files(files)
```

frankenjoe:
Closes #65 #25 

I decided to apply a technique similar to how we pass `win_dur` and `hop_dur` to the processing function of `Feature`:

image

This has the advantage that we do not have to add a new argument. Special arguments are `'idx'` (running index), `'file'` (current file) and `'root'` (root folder)

### Usage example

image

### Example

```python
db = audb.load(
    'emodb',
    version='1.2.0',
    format='wav',
    mixdown=True,
    sampling_rate=16000,
    full_path=False,
)


def process_func(signal, sampling_rate, idx):
    return idx


interface = audinterface.Process(
    process_func=process_func,
    num_workers=10,
)
interface.process_files(db.files, root=db.root)
```
```
file             start   end                      
wav/03a01Fa.wav  0 days  0 days 00:00:01.898250         0
wav/03a01Nc.wav  0 days  0 days 00:00:01.611250         1
wav/03a01Wa.wav  0 days  0 days 00:00:01.877812500      2
wav/03a02Fc.wav  0 days  0 days 00:00:02.006250         3
wav/03a02Nc.wav  0 days  0 days 00:00:01.439812500      4
                                                     ... 
wav/16b10Lb.wav  0 days  0 days 00:00:03.442687500    530
wav/16b10Tb.wav  0 days  0 days 00:00:03.500625       531
wav/16b10Td.wav  0 days  0 days 00:00:03.934187500    532
wav/16b10Wa.wav  0 days  0 days 00:00:02.414125       533
wav/16b10Wb.wav  0 days  0 days 00:00:02.522499999    534
```

If used with a sliding window, the index is repeated for frames belonging to the same file:

```python
interface = audinterface.Process(
    process_func=process_func,
    num_workers=10,
    win_dur=1.0,
)
y = interface.process_files(db.files, root=db.root)
```
```
file             start                   end                   
wav/03a01Fa.wav  0 days 00:00:00         0 days 00:00:01             0
                 0 days 00:00:00.500000  0 days 00:00:01.500000      0
wav/03a01Nc.wav  0 days 00:00:00         0 days 00:00:01             1
                 0 days 00:00:00.500000  0 days 00:00:01.500000      1
wav/03a01Wa.wav  0 days 00:00:00         0 days 00:00:01             2
                                                                  ... 
wav/16b10Wa.wav  0 days 00:00:01         0 days 00:00:02           533
wav/16b10Wb.wav  0 days 00:00:00         0 days 00:00:01           534
                 0 days 00:00:00.500000  0 days 00:00:01.500000    534
                 0 days 00:00:01         0 days 00:00:02           534
                 0 days 00:00:01.500000  0 days 00:00:02.500000    534
```

Also works with other interface classes:

```python
    'index',
    process_func=process_func,
    num_workers=10,
)
interface.process_files(db.files, root=db.root)
```
```
                                                  index
file            start  end                             
wav/03a01Fa.wav 0 days 0 days 00:00:01.898250         0
wav/03a01Nc.wav 0 days 0 days 00:00:01.611250         1
wav/03a01Wa.wav 0 days 0 days 00:00:01.877812500      2
wav/03a02Fc.wav 0 days 0 days 00:00:02.006250         3
wav/03a02Nc.wav 0 days 0 days 00:00:01.439812500      4
...                                                 ...
wav/16b10Lb.wav 0 days 0 days 00:00:03.442687500    530
wav/16b10Tb.wav 0 days 0 days 00:00:03.500625       531
wav/16b10Td.wav 0 days 0 days 00:00:03.934187500    532
wav/16b10Wa.wav 0 days 0 days 00:00:02.414125       533
wav/16b10Wb.wav 0 days 0 days 00:00:02.522499999    534
```

And `ProcessWithContext`:

```python
def process_func(signal, sampling_rate, starts_i, ends_i, idx, file):
    return [(idx, file)] * len(starts_i)


interface = audinterface.ProcessWithContext(
    process_func=process_func,
)

index = audformat.segmented_index(
    [db.files[0], db.files[0], db.files[1]],
    [0, 1, 0],
    [1, 2, 1],
)
interface.process_index(index, root=db.root)
```
```
file             start            end            
wav/03a01Fa.wav  0 days 00:00:00  0 days 00:00:01    (0, wav/03a01Fa.wav)
                 0 days 00:00:01  0 days 00:00:02    (0, wav/03a01Fa.wav)
wav/03a01Nc.wav  0 days 00:00:00  0 days 00:00:01    (1, wav/03a01Nc.wav)
dtype: object
```


frankenjoe:
@hagenw I put WIP as the implementation with `ProcessWithContext` is still missing and we should probably also provide a usage example. But I wanted to get your feedback on the proposed solution first.

hagenw:
It is slightly strange with the current implementation that it returns a single index value instead of the actual index value. This means inside the processing function you seem not to have access to the actual index values?

frankenjoe:
> It is slightly strange with the current implementation that it returns a single index value instead of the actual index value

Sorry, don't get what you mean by this sentence.

frankenjoe:
Ah, do you mean with actual e.g. the file? Or the segment? I think it's easier this way as it's always an integer and does not depend on the processing function that is called.

hagenw:
```python
def process_func(signal, sampling_rate, idx):
    return idx


interface = audinterface.Process(
    process_func=process_func,
    num_workers=10,
)
interface.process_files(db.files, root=db.root)
```
For this I would have expected something like:

```
file             start                   end                   
wav/03a01Fa.wav  0 days 00:00:00         0 days 00:00:01             (wav/03a01Fa.wav , 0 days 00:00:00, 0 days 00:00:01)
...
```
but with `pd.Timedelta` objects not the stuff I added.

frankenjoe:
I think it's easier to only get an index. If you need the file you can do:

```
def process_func(signal, sampling_rate, idx, files):
    return files[idx]


interface = audinterface.Process(
    process_func=process_func,
    process_func_args={'files': db.files},
    num_workers=10,
)
interface.process_files(db.files, root=db.root)
```

or simply:

```python
def process_func(signal, sampling_rate, idx):
    return db.files[idx]
```

hagenw:
> Ah, do you mean with actual e.g. the file or the index segment? I think it's easier this way as it's always an integer and does not depend on the processing function that is called.

I agree that it is easier and avoids the hassle of returning strange objects in the dataframe.

I guess you can still access the actual index entries then with:

```python
db = audb.load(...)
index = db['files'].files

def process_func(signal, sampling_rate, idx):
   end = index[idx].get_level_values('start')
```
?

frankenjoe:
> I guess you can still access the actual index entries then with:
> 
> ```python
> db = audb.load(...)
> index = db['files'].files
> 
> def process_func(signal, sampling_rate, idx):
>    end = index.get_level_values('start')[idx]
> ```
> 
> ?

Yes, however, since `end` can be affected by `min_signal_duration` and `max_signal_duration` it may not match the value in the index of the returned series.

I now implemented also `file` and `root` as special arguments. The question is if we should also support `start` and `end`. I already gave it a try, but it didn't work out of the box and will require some refactoring of the logic. So the question is if we really need it?

hagenw:
I would stay with the running index and not provide a way to get `start` or `end`. If the user really needs this, she can add `index` as another argument to the processing function as she would do for gender.

hagenw:
> I now implemented also file and root as special arguments

What does this mean?

frankenjoe:
> What does this mean?

E.g.:

```python
...

def process_func(signal, sampling_rate, idx, root, file):
    return (idx, file, root)


interface = audinterface.Process(process_func=process_func)
interface.process_files(db.files, root=db.root)
```
```
Get:   emodb v1.2.0
Cache: /media/jwagner/Data/audb/emodb/1.2.0/fe182b91
file             start   end                      
wav/03a01Fa.wav  0 days  0 days 00:00:01.898250       (0, wav/03a01Fa.wav, /media/jwagner/Data/audb/...
wav/03a01Nc.wav  0 days  0 days 00:00:01.611250       (1, wav/03a01Nc.wav, /media/jwagner/Data/audb/...
wav/03a01Wa.wav  0 days  0 days 00:00:01.877812500    (2, wav/03a01Wa.wav, /media/jwagner/Data/audb/...
wav/03a02Fc.wav  0 days  0 days 00:00:02.006250       (3, wav/03a02Fc.wav, /media/jwagner/Data/audb/...
wav/03a02Nc.wav  0 days  0 days 00:00:01.439812500    (4, wav/03a02Nc.wav, /media/jwagner/Data/audb/...
                                                                            ...                        
wav/16b10Lb.wav  0 days  0 days 00:00:03.442687500    (530, wav/16b10Lb.wav, /media/jwagner/Data/aud...
wav/16b10Tb.wav  0 days  0 days 00:00:03.500625       (531, wav/16b10Tb.wav, /media/jwagner/Data/aud...
wav/16b10Td.wav  0 days  0 days 00:00:03.934187500    (532, wav/16b10Td.wav, /media/jwagner/Data/aud...
wav/16b10Wa.wav  0 days  0 days 00:00:02.414125       (533, wav/16b10Wa.wav, /media/jwagner/Data/aud...
wav/16b10Wb.wav  0 days  0 days 00:00:02.522499999    (534, wav/16b10Wb.wav, /media/jwagner/Data/aud...
Length: 535, dtype: object
```

hagenw:
OK, I see.

It might be handy to have, but I'm not sure if it is neccessary. They make not sense in all cases, e.g. `process_signal()` and there are workarounds to get that information:

```python
def process_func(signal, sampling_rate, idx, index, root):
    if audformat.index_type == 'filewise':
        file = str(index[idx])
    else:
        file = index.get_level_values('file')[idx]
    return (idx, file, root)

interface = audinterface.Process(
    process_func=process_func,
    process_func_args={
        'index': db.files,
        'root': db.root,
    },
)
```

So maybe postpone `file` and `root` to another pull request and see if we really need them?

There is also a slight difference between `file` and `root` as `root` you can/must provide to `process_file()`, `process_files()`, `process_folder()`, `process_index()`. So maybe in the case of `root` it makes sense to add it.

frankenjoe:
> They make not sense in all cases, e.g. `process_signal()`

Not necessarily:

```python
sampling_rate = 8000
signal = np.zeros(sampling_rate, np.float32)
interface.process_signal(signal, sampling_rate, file='file.wav')
```
```
file      start   end            
file.wav  0 days  0 days 00:00:01    (0, file.wav, None)
dtype: object
```

If `file` or `root` are not provided they will be set to `None`.

I would be in favor of keeping `file` since one of the use cases I see is the following:

```
def process_func(signal, sampling_rate, file, cache_root):
    y = ...
    path = audeer.path(cache_root, file)
    y.to_pickle(path)
    return path
```

hagenw:
OK, then let's continue here with `root` and `file`.

frankenjoe:
I also implemented special argument for `ProcessWithContext` and added a usage example (see description). This is now ready for review.

Do you think we can close #25 with this PR as well?

hagenw:
And you even provide an example (for gender) in the usage section.

tests/test_process.py
```diff
@@ -1283,6 +1283,128 @@ def test_process_with_sliding_window(
...
+            [0] * num_files,
+            [duration] * num_files,
+        ),
+    )
+    pd.testing.assert_series_equal(y, expected)
+
+    # process index
+    y = process.process_index(index, root=root)
+    expected = pd.Series(
+        [(idx, file, root) for idx, (file, _, _) in enumerate(index)],
```
hagenw, line 1356: Would be nice to also add a test for a filewise index

audinterface/core/process.py
```diff
@@ -165,6 +177,21 @@ def process_func(signal, _):
...
+            'idx': False,
+            'root': False,
+            'file': False,
+        }
+        for key in self._process_func_special_args:
+            if (
+                    key in signature.parameters
+                    and key not in process_func_args
+            ):
+                self._process_func_special_args[key] = True
```
hagenw, line 193: As you need to repeat this for each interface and it is not so easy to read, maybe replace it by a helper function?

Then you could write here:

```python
self._process_func_special_args = utils.check_special_args(
    ['idx', 'root', 'file'],
    process_func,
    process_func_args,
)
```

The only downside it that you not easily see then that a dictionary is created.

frankenjoe: Actually, not every interface, just `Process' and once for `ProcessWithContext`. So I'm not sure if it's really worth to create a helper function for it.

tests/test_process.py
frankenjoe: done

audinterface/core/process.py
hagenw: OK, and as you can more easily see that we assign dictionaries, we simply stay with the current solution.

PR #69, opened 2022-07-14
TST: test for audobject

### LINKED ISSUE #68: Error when using an audonnx object as processing function
As reported in https://github.com/audeering/w2v2-how-to/issues/13 the newest version of `audinterface` seems no longer compatible with `audonnx`.

Try the following code from https://github.com/audeering/w2v2-how-to/blob/master/notebook.ipynb

```python
import os

import audeer
import audinterface
import audonnx


model_root = 'model'
cache_root = 'cache'


audeer.mkdir(cache_root)
def cache_path(file):
    return os.path.join(cache_root, file)


url = 'https://zenodo.org/record/6221127/files/w2v2-L-robust-12.6bc4a7fd-1.1.0.zip'
dst_path = cache_path('model.zip')

if not os.path.exists(dst_path):
    audeer.download_url(
        url, 
        dst_path, 
        verbose=True,
    )
    
if not os.path.exists(model_root):
    audeer.extract_archive(
        dst_path, 
        model_root, 
        verbose=True,
    )

model = audonnx.load(model_root)

sampling_rate = 16000
    model.outputs['logits'].labels,
    process_func=model,
    process_func_args={
        'output_names': 'logits',
    },
    sampling_rate=sampling_rate,
    resample=True,    
    verbose=True,
)
```

This results in

```
AttributeError                            Traceback (most recent call last)
<ipython-input-7-447aa9996e02> in <module>                                                  
----> 1 interface = audinterface.Feature(
      2     model.outputs['logits'].labels,           
      3     process_func=model,           
      4     process_func_args={
      5         'output_names': 'logits',

~/git/audeering/audinterface/audinterface/core/feature.py in new_func(*args, **kwargs)
     44                 )
     45                 warnings.warn(message, category=UserWarning, stacklevel=2)
---> 46             return func(*args, **kwargs)
     47         return new_func
     48     return _deprecated

~/git/audeering/audinterface/audinterface/core/feature.py in __init__(self, feature_names, name, params, process_func, process_func_args, process_func_is_mono, process_func_applies_sliding_wi
ndow, sampling_rate, resample, channels, mixdown, win_dur, hop_dur, min_signal_dur, max_signal_dur, segment, keep_nat, num_workers, multiprocessing, verbose, **kwargs)
    312         # add 'win_dur' and 'hop_dur' to process_func_args
    313         # if expected by function but not yet set
--> 314         signature = inspect.signature(process_func)
    315         if (
    316             'win_dur' in signature.parameters

/usr/lib/python3.8/inspect.py in signature(obj, follow_wrapped)
   3103 def signature(obj, *, follow_wrapped=True):
   3104     """Get a signature object for the passed callable."""
-> 3105     return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
   3106
   3107

/usr/lib/python3.8/inspect.py in from_callable(cls, obj, follow_wrapped)
   2852     def from_callable(cls, obj, *, follow_wrapped=True):
   2853         """Constructs Signature for the given callable object."""
-> 2854         return _signature_from_callable(obj, sigcls=cls,
   2855                                         follow_wrapper_chains=follow_wrapped)
   2856

/usr/lib/python3.8/inspect.py in _signature_from_callable(obj, follow_wrapper_chains, skip_bound_arg, sigcls)
   2305                                         skip_bound_arg=skip_bound_arg)
   2306
-> 2307     if _signature_is_builtin(obj):
   2308         return _signature_from_builtin(sigcls, obj,
   2309                                        skip_bound_arg=skip_bound_arg)

/usr/lib/python3.8/inspect.py in _signature_is_builtin(obj)
   1845             # Can't test 'isinstance(type)' here, as it would
   1846             # also be True for regular python classes
-> 1847             obj in (type, object))
   1848
   1849

~/.envs/analysis/lib/python3.8/site-packages/audobject/core/object.py in __eq__(self, other)
    405
    406     def __eq__(self, other: 'Object') -> bool:
--> 407         return self.id == other.id
    408
    409     def __repr__(self) -> str:

AttributeError: type object 'type' has no attribute 'id'
```

hagenw:
Closes #68 

This closes the issue by adding a test for `audobject` to identify possible issues earlier.

The actual issue was resolved in `audobject`: in https://github.com/audeering/audobject/pull/49

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,6 @@ classifiers =
 packages = find:
 install_requires =
     audformat >=0.15.3,<2.0.0
-    audobject >=0.7.2
     audresample >=1.1.0,<2.0.0
 python_requires = >=3.8
 setup_requires =
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,6 +27,7 @@ classifiers =
 packages = find:
 install_requires =
     audformat >=0.15.3,<2.0.0
+    audobject >=0.7.2
     audresample >=1.1.0,<2.0.0
 python_requires = >=3.8
 setup_requires =
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,3 +1,4 @@
+audobject >=0.7.5
 # To avoid https://github.com/tholo/pytest-flake8/issues/87
 flake8 <5.0.0
 pytest
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -8,6 +8,7 @@ import pytest
 
 import audinterface
 import audformat
+import audobject
 
 
 def signal_duration(signal, sampling_rate):
@@ -16,6 +17,12 @@ def signal_duration(signal, sampling_rate):
 
 def signal_max(signal, sampling_rate):
     return np.max(signal)
+
+
+class SignalObject(audobject.Object):
+
+    def __call__(self, signal, sampling_rate):
+        return np.max(signal)
 
 
 SEGMENT = audinterface.Segment(
@@ -968,6 +975,12 @@ def test_process_signal(
         ),
         (
             signal_max,
+            np.random.random(5 * 44100),
+            44100,
+            audinterface.utils.signal_index(),
+        ),
+        (
+            SignalObject(),
             np.random.random(5 * 44100),
             44100,
             audinterface.utils.signal_index(),


frankenjoe:
I think so far we did not have a dependency to `audobject`, so I'm not sure if we should really introduce one here.

frankenjoe:
On the other hand it might be easier than adding it to all packages that actually rely on `audinterface` and `audobject`.

hagenw:
I will update this pull request remove the dependency on `audobject`, but add a test for it to the tests.

frankenjoe:
> Ah, you are right, we do not need `audobject` here. So, I would also not add it, but we should updat ethe `audonnx` package instead.

Yes, I think that's the better approach. I guess we should also update the `opensmile` package.

frankenjoe:
> Yes, I think that's the better approach. I guess we should also update the `opensmile` package.

Turns out, we don't since there we are not relying on `win_dur` and `hop_dur`, since we simply apply the frame times that we receive from the C library.

hagenw:
I added a test for `audobject` and updated the description.

PR #73, opened 2022-10-06
DOC: fix return values of utils.read_audio()

hagenw:
Changes

image

to

image

It seems using `signal: ` before had the consequence that this was used as return type and even affected the order in which "Return type" and "Returns" are presented. 
The order shown for the changes here is in line with the other docstrings.


hagenw:
https://github.com/audeering/audinterface/pull/75 needs to be merged first

PR #74, opened 2022-10-06
DOC: add missing return type to sliding_window()

hagenw:
The return type of `audinterface.utils.sliding_window()` was missing, now we have

image


hagenw:
https://github.com/audeering/audinterface/pull/75 needs to be merged first

PR #75, opened 2022-10-06
Require audformat>=0.15.3 and flake8<5.0.0

hagenw:
This pull requests solves two issues to get the tests working again:

### 1. segmented indices and new `string` dtype

It makes sure to use `audformat.segmented_index()` instead of `pd.MultiIndex` and `audformat.utils.union()` instead of `index.union()` whenever dealing with segemented indices to make sure we get the correct dtype of the `file` level. It uses now only `audformat` functions to make sure it will still work even if we change the behavior of `audformat.sefgmented_index()` again.

### 2. remove support for Python 3.7

As `audformat>=0.15` does only work with Python 3.8, I removed support for Python 3.7 as well.

### 3. flake8 bug

Seems like they don't want to solve https://github.com/tholo/pytest-flake8/issues/87, but propose instead of doing syntax/style checks as `pre-commit`. If I understand this correctly it would run `flake8` locally before doing the commit.
The disadvantage would be that we wouldn't know if everyone uses the same flake8 version.

So for now, let's also restrict the requirements.

### DIFF
--- a/audinterface/core/segment.py
+++ b/audinterface/core/segment.py
@@ -532,16 +532,7 @@ class Segment:
             progress_bar=self.process.verbose,
             task_description=f'Process {len(index)} segments',
         )
-
-        if len(index.levels) == 2:
-            # Index without 'file' level
-            index = y[0]
-            for obj in y[1:]:
-                index = index.union(obj)
-        else:
-            # For segmented index use audformat.utils.union()
-            # to ensure the correct dtypes are preserved
-            index = audformat.utils.union(y)
+        index = audformat.utils.union(y)
 
         return index
 
--- a/audinterface/core/segment.py
+++ b/audinterface/core/segment.py
@@ -284,18 +284,10 @@ class Segment:
             end=end,
             root=root,
         ).values[0]
-        return pd.MultiIndex(
-            levels=[
-                [file],
-                index.levels[0] + start,
-                index.levels[1] + start,
-            ],
-            codes=[
-                [0] * len(index),
-                index.codes[0],
-                index.codes[1],
-            ],
-            names=['file', 'start', 'end'],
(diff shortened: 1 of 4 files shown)


hagenw:
Oh, seems like we have also not updated the tests for the new `audformat` version.
I will take a look.

audinterface/core/segment.py
```diff
@@ -555,9 +533,15 @@ def process_signal_from_index(
...
-            index = index.union(obj)
+        if len(index.levels) == 2:
+            # Index without 'file' level
+            index = y[0]
+            for obj in y[1:]:
+                index = index.union(obj)
+        else:
+            # For segmented index use audformat.utils.union()
+            # to ensure the correct dtypes are preserved
+            index = audformat.utils.union(y)
```
hagenw, line 544: If we would require `audformat>=0.15.3` we could reduce this further to just
```python
        index = audformat.utils.union(y)
```

Would there be any downsides in asking for a new `audformat` version?

audinterface/core/segment.py
```diff
@@ -338,21 +330,15 @@ def process_files(
             ends=ends,
             root=root,
         )
+        if len(series) == 0:
+            return audformat.filewise_index()
```
hagenw, line 334: I had to add this for `audformat>=0.15.0` as `audformat.utils.union([])` now returns:
```
Index([], dtype='object')
```

audinterface/core/segment.py
frankenjoe: I don't think so, let's simplify.

hagenw: As the new audformat only works with Python >=3.8, I also had to remove support for Python 3.7.

PR #78, opened 2022-10-20
Use multi-column format for multi-channel Feature

### LINKED ISSUE #77: Use multi index columns when extracting features from multiple channels
When extracting features from multiple channels, we currently expand the feature name by the channel number, e.g.:

```python
    ['min', 'max'],
    process_func=lambda x, sr: [x.min(axis=1), x.max(axis=1)],
    channels=[0, 1],
)

df = interface.process_file('stereo.wav')
df
```
```
                                                min-0    max-0     min-1     max-1
file       start  end                                                             
stereo.wav 0 days 0 days 00:00:07.247709751 -0.723999 -0.72403  0.782379  0.782349
```

Hence, there is not straight-forward way for a user to select the features belonging to a specific channel.

A more convenient way might be the use of `MultiIndex` columns, e.g. the previous `DataFrame` could be converted to:

```python
dfmi = pd.DataFrame(
    df.values,
    columns=pd.MultiIndex.from_tuples(
        [
            (0, 'min'),
            (0, 'max'),
            (1, 'min'),
            (1, 'max'),
        ]
    ),
    index=df.index,
)
dfmi
```
```
                                                    0                  1          
                                                  min      max       min       max
file       start  end                                                             
stereo.wav 0 days 0 days 00:00:07.247709751 -0.723999 -0.72403  0.782379  0.782349
```

Return features of second channel:

```python
dfmi[1]
```
```
                                                  min       max
file       start  end                                          
stereo.wav 0 days 0 days 00:00:07.247709751  0.782379  0.782349
```

Access single feature:

```python
dfmi[1]['min']
```
```
file        start   end                      
stereo.wav  0 days  0 days 00:00:07.247709751    0.782379
```

frankenjoe:
Closes #77 

### API

image
...
image

### Usage

image

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -280,10 +280,11 @@ class Feature:
             column_names = []
             for channel in range(num_channels):
                 column_names.extend(
-                    [f'{name}-{channel}' for name in feature_names]
+                    [(channel, feature_name) for feature_name in feature_names]
                 )
+            column_names = pd.MultiIndex.from_tuples(column_names)
         else:
-            column_names = feature_names
+            column_names = pd.Index(feature_names)
 
         if process_func is None:
             def process_func(signal, _):
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -83,9 +83,11 @@ class Feature:
 
     Args:
         feature_names: features are stored as columns in a data frame,
-            ``feature_names`` defines the names of the columns.
-            If ``num_channels`` > 1,
-            the channel number will be appended to the column names.
+            where ``feature_names`` defines the names of the columns.
+            If ``len(channels)`` > 1,
+            the data frame has a multi-column index with
+            with channel ID as first level
+            and ``feature_names`` as second level
         name: name of the feature set, e.g. ``'stft'``
         params: parameters that describe the feature set,
             e.g. ``{'win_size': 512, 'hop_size': 256, 'num_fft': 512}``.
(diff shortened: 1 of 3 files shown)


frankenjoe:
As already mentioned in #77 this introduces a breaking change. But I guess that's ok, we anyway had not covered the usage of multi-channel input in the usage section yet.

tests/test_feature.py
```diff
@@ -438,9 +438,75 @@ def process_func(s, sr, arg1, arg2):
         )
 
 
+def test_process_index(tmpdir):
```
frankenjoe, line 441: I just moved this function to restore alphabetical order.

hagenw:
> As already mentioned in #77 this introduces a breaking change. But I guess that's ok, we anyway had not covered the usage of multi-channel input in the usage section yet.

I guess we should figure out a way to scan our code for cases where this change would have an effect?

audinterface/core/feature.py
```diff
@@ -278,12 +305,13 @@ def add_unit(dur, unit):
         feature_names = audeer.to_list(feature_names)
         if num_channels > 1:
             column_names = []
-            for channel in range(num_channels):
```
frankenjoe, line 281: So far we had always named the channels `0, 1, ...`, but I think it makes more sense to stick to the actual channel IDs. That way you will later see directly from the column names which channels of the file were processed.

docs/usage.rst
```diff
@@ -21,7 +21,7 @@
                 level=0,
             )
 
-        return df.to_html(max_rows=6, max_cols=3)
+        return df.to_html(max_rows=6, max_cols=4)
```
hagenw, line 24: I guess this change is responsible for changing

image

to

image

?

So maybe we switch to use `db.root` to make the `file` entry shorter as you suggested some time ago?

frankenjoe: Yes, I wanted to make sure that we see all columns in the multi-channel example:

image

> So maybe we switch to use db.root to make the file entry shorter as you suggested some time ago?

I think we can achieve the same with the current solution, we simply remove a little more.

frankenjoe: I changed it to:

image

hagenw: Great, thanks.

Issue #79, opened 2022-10-21
TOC does now show automatic generated sub-sections


hagenw:
![image](https://user-images.githubusercontent.com/173624/197133371-56d26176-22b8-4a27-a254-45f56edd5068.png)

I can reproduce this locally when upgrading to the latest `sphinx` version.
So i guess this is a new feature.

hagenw:
It was introduced in https://github.com/sphinx-doc/sphinx/issues/6316 (sphinx 5.2.0), and after complains they also added a config setting to be able to disable it:  `nocontentsentry` (sphinx 5.2.3).

I would propose for the short term to disable it.

For the long-term we could use it to generate better TOCs as I always wanted to add a way to include the possibility to select a method in the TOC. The way I had in mind was to change to a structure used by `pandas` and `numpy` with having single pages for the classes, e.g. https://pandas.pydata.org/docs/reference/api/pandas.Series.html but the downside wis that this would break all existing links. So maybe we could use this new feature to work around this and stay with our current structure.

PR #80, opened 2022-10-21
DOC: disable auto-listing of API entries

hagenw:
First workaround for #79 

This disables the automatic generated entries to restore the original API TOC:

image


frankenjoe:
Can't really spot any difference, what exactly is fixed here?

hagenw:
In the latest published version it looks like this:

image

frankenjoe:
Strange, for me it already looks with the current master:

image

frankenjoe:
Ah, after updating `sphinx` from `5.2.0` to `5.3.0` I now see a difference :)

PR #84, opened 2022-10-24
DOC: fix typo in multi-channel usage example

hagenw:
Before we were passing the mono signal to the first example showing that the first channel is used.

Fixed now:

image


frankenjoe:
Thanks for spotting, example makes more sense now :)

PR #85, opened 2022-12-14
WIP: DOC: split into sub-pages and use autosummary

hagenw:
This introduces a new structure for our documentation, that I would like to propose for all our packages.
To avoid very long pages and provide also links for class methods and attributes in the TOC, it uses [sphinx.ext.autosummary](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html) to automatically split up the documentation into sub-pages.

All changes are explained in detail in the following sub-sections.

## autosummary and templates

`sphinx.ext.autosummary` is able to create TOC entries and corresponding sub-pages (which means RST files that also are stored within `docs/api/` when used with the `:toctree:` option.
What exactly is included in pages that are automatically generated can then be configured by templates. The important one here is the `docs/_templates/autosummary/class.rst` in which we define how a sub-page for a class should be presented. I opted to list methods (including properties) and attributes in the same table in order to have them alphabetically ordered in the TOC.
You could also present them in different sections as `pandas` is doing it, compare https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.html.

The definition of the templates could also be done in the theme, but it might be that we want to adapt them inside projects, e.g. here we want to list the special method `__call__`, but in other projects we might want to not list it.
So for now I would propose we include the templates in the single projects.

## Predictable URL structure

The URL of a given class/function/method etc. is completely given by the module name and all the other names, e.g.

| Name | URL |
| --- | --- |
| `audinterface` | `base-url/api/audinterface.html` |
| `audinterface.Feature` | `base-url/api/audinterface.Feature.html` |
| `audinterface.Feature.process_index()` | `base-url/api/audinterface.Feature.process_index.html` |
| `audinterface.Feature.name` | `base-url/api/audinterface.Feature.name.html` |
| `audinterface.utils` | `base-url/api/audinterface.utils.html` |
| `audinterface.utils.read_audio()` | `base-url/api/audinterface.utils.read_audio.html` |

The downside is that this will break existing links to the API in other published documentations.
For that reason it would be my goal to update all our API documentations with this new schema until mid of January, then we should be able to eliminate all broken links.

## Inclusion in docs via (sub)-module files

As before we have to create entry files for each (sub)-module, but those now have an easier structure,
e.g. `docs/api/audinterface.utils.rst`:

```rst
audinterface.utils
==================

.. automodule:: audinterface.utils

.. autosummary::
    :toctree:
    :nosignatures:

    read_audio
    signal_index
    sliding_window
    to_timedelta
```

Which will result in the following rendered page:

image

showing the following TOC entry

image

## Listing of all class methods/properties/attributes

We now get single pages for each class, but also for all the methods and attributes of a class.
The main page of a class does now simply have a table at the bottom under the "Examples" section  that lists everything that is available and shows it also in the TOC entry, e.g.

image

image

**Note:** even though I sort all entries alphabetically in the corresponding template `docs/_templates/autosummary/class.rst`, there is still an error in the order. `process_func` and `process_func_args` should be listed before `process_signal_from_index`. I need to figure out how to fix this.

## Updates to Example section

* Rename to "Examples"
* Split up into several blocks as done in `pandas` and `numpy`, e.g. https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.abs.html#pandas.Series.abs

image

**Note:** it looks slightly strange as we add left padding to the code block in the example sections with our theme. This is not the default behavior, and code blocks in the example sections of `pandas` and `numpy` are not padded on the left. So we might think to remove the extra padding in our theme.

## Theme outlook

I'm also planning to adjust our `sphinx-audeering-theme` to make it fit with API documentation presented on a single page as we did before (and as maybe other users would like to continue, e.g. for C projects) and presented on several pages with `autosummary`. This is not ready and will take some time to finish it properly, but just to provide an outlook how things might look like I add a screenshot here:

image


frankenjoe:
Cool, looking forward to see the proposed changes in action. Providing links to class methods etc. will be extremely useful.

hagenw:
> Cool, looking forward to see the proposed changes in action. Providing links to class methods etc. will be extremely useful.

You can provide links already with the current documentation, e.g. https://audeering.github.io/audinterface/api.html#audinterface.Process.process_index, but we cannot easily list them in the TOC and the page also becomes very crowded.

This should be ready for review tomorrow.

hagenw:
The overall implementation and the description in the pull request is now finished, you might already check the changes out, build the documentation and have a look.

At the moment there is still a small bug regarding the alphabetical sorting of class methods and attributes, but besides this it should be ready.

frankenjoe:
At a first glance it looks great already. I have two remarks, though:

1. I think it would be useful if one could immediately see what are class attributes / properties and what are methods. Maybe by adding `()` to the methods? Maybe it also makes sense to first list the attributes / properties and afterwards the methods.
2. It might be useful to still show a single page with all attributes / methods of a class and only jump to the according anchor when clicking on an attribute / method in the TOC. Reason: sometimes it can be useful to quickly search for a keyword across all attributes / methods.

hagenw:
> Maybe it also makes sense to first list the attributes / properties and afterwards the methods.

This was also my first idea how to approach this, but I run into a few problems with it:

* I didn't find a way to present methods and attributes in different tables, but still list them alphabetically in the TOC. I would very much like an alphabetically listing there as a user might not know if the stuff she/he is looking for is a method or attribute. An example for first listing attributes and then methods can also be find in https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.html. If you look at the left TOC you will see that it first lists all attributes alphabetically and afterwards all methods
* I didn't find a way so far to differentiate between methods and properties

hagenw:
> Maybe by adding () to the methods?

That's a good idea, I will try to achieve this. By not using `:nosignatures:` one could list the whole function signature, but this looks much to crowded in my opinion. Just adding `()` would be perfect

hagenw:
> It might be useful to still show a single page with all attributes / methods of a class and only jump to the according anchor when clicking on an attribute / method in the TOC. Reason: sometimes it can be useful to quickly search for a keyword across all attributes / methods.

Search is indeed the biggest thing negatively affected by the current approach. As a workaround you can use the build-in sphinx search, which works reasonable well.

But as a comparison I will also try to create another pull request that implements your solution.
One workaround we will need there is to add headings for each attribute and method in order to make them appear in the TOC.

hagenw:
> > Maybe it also makes sense to first list the attributes / properties and afterwards the methods.
> 
> This was also my first idea how to approach this, but I run into a few problems with it:
> 
> * I didn't find a way to present methods and attributes in different tables, but still list them alphabetically in the TOC. I would very much like an alphabetically listing there as a user might not know if the stuff she/he is looking for is a method or attribute. An example for first listing attributes and then methods can also be find in https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.html. If you look at the left TOC you will see that it first lists all attributes alphabetically and afterwards all methods
> * I didn't find a way so far to differentiate between methods and properties

Good news, I found now a way to separate the presentation of the methods/attributes in the summary table on the class page and the order of the corresponding entries in the TOC.

So now we have:

image

and the corresponding TOC entry

image

hagenw:
> > > Maybe it also makes sense to first list the attributes / properties and afterwards the methods.
> > 
> > 
> > This was also my first idea how to approach this, but I run into a few problems with it:
> > 
> > * I didn't find a way to present methods and attributes in different tables, but still list them alphabetically in the TOC. I would very much like an alphabetically listing there as a user might not know if the stuff she/he is looking for is a method or attribute. An example for first listing attributes and then methods can also be find in https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.html. If you look at the left TOC you will see that it first lists all attributes alphabetically and afterwards all methods
> > * I didn't find a way so far to differentiate between methods and properties
> 
> Good news, I found now a way to separate the presentation of the methods/attributes in the summary table on the class page and the order of the corresponding entries in the TOC.
> 
> So now we have:
> 
> image
> 
> and the corresponding TOC entry
> 
> image
> 
> The problem with the properties as part of the methods still exist, but inside `audinterface` we don't have any properties and we will have this problem only in other packages at the moment.

Unfortunately it is not as easy as I have thought.

You can only generate stub files (that's the name of the files that need to be auto-generated for each method/attribute) if you use the `:toctree:` option with `.. autosummary::`, which means those files are not generated with my current approach.

If I add the `:toctree:` argument back there is no way to hide the entries in the TOC (at least no easy way, maybe we can find some manual tweaks) and we would end up with repeated entries (one time not ordered, the time ordered alphabetically):

image

hagenw:
I now found a way to auto-generate stub files and still be able to define the presentation order in the TOC independently from the presentation order on the class page. It's of cause not so nice and more of a hack, but maybe we can continue from here and see if we can improve the approach. The outcome looks exactly as first presented in https://github.com/audeering/audinterface/pull/85#issuecomment-1353016772

The downside is that we now need two different template versions and two commands to build the documentation.
We first create the stub files with the templates stored inside `docs/_templates-stub-files/`:

```bash
$ sphinx-autogen -o docs/api/ -t docs/_templates-stub-files/ docs/api/*.rst
```

Afterwards we build the documentation using the `docs/_templates` folder as specified inside `docs/conf.py`, which will not generate any stub files and add custom TOC entries with our usual command:

```bash
$ python -m sphinx docs/ build/html -b html
```

This works great within an CI pipeline, but it has the big downside that a user now would have to run two commands to build the documentation.

hagenw:
We could avoid the need to run two commands, by adding a command to `docs/conf.py` as we could import from https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autosummary/generate.py.

I will in general have a closer look at `autosummary` to see if it would make more sense to have some custom code or create an issue/pull request at sphinx or life with some options from `autosummary` to get the desired results.

hagenw:
I added now custom code to `docs/conf.py` to generate the stub files. It also has the advantage that we now store only generated files under `docs/api` and have source RST files stored under `docs/api-src`.

hagenw:
I created now a version that shows a single page for classes containing all methods and attributes to make it easier search the page at https://github.com/audeering/audinterface/pull/86.

One advantage is that we don't have to handle the complicated `autosummary` two run scenario that I used here to get the correct order in the TOC.

The downside of a single page is that it can still be crowded if you have a lot of methods and all come with long examples.

frankenjoe:
> Good news, I found now a way to separate the presentation of the methods/attributes in the summary table on the class page and the order of the corresponding entries in the TOC.

Now that we seperate methods and attributes in the summary table, wouldn't it make sense to show the same order in the TOC? I.e. first the methods ordered alphabetically and then the attributes ordered alphabetically. I think that's what they also do in the pandas documentation.

hagenw:
> > Good news, I found now a way to separate the presentation of the methods/attributes in the summary table on the class page and the order of the corresponding entries in the TOC.
> 
> Now that we separate methods and attributes in the summary table, wouldn't it make sense to stick to the same order in the TOC? I.e. first the methods ordered alphabetically and then the attributes ordered alphabetically. I think that's what they also do in the pandas documentation.

Yes, but I don't like it very much, because the user has to now beforehand if the name belongs to a method or an attribute. And if we split methods and attributes, we would need to split properties as well in my opinion and then you will have three different alphabetically ordered lists in the TOC.

frankenjoe:
Mh ok, but it would be really helpful do quickly identify the methods, cause this is what in 95% of the cases I'm looking for when I want to learn how to interact with a class. Adding a `()` to the methods in the TOCs is not possible, right?  Or maybe adding a blank link between methods and attributes?

hagenw:
That's a problem I was not able to solve with `autosummary`. So far I didn't found a way to add `()` to the TOC entries or the entries in the tables created by autosummary.

In https://github.com/audeering/audinterface/pull/86 we could add `()` to the headings, which would add those also to the TOC entries.

hagenw:
I updated the example in #86 and added `()` at the end of functions and methods.

frankenjoe:
Looks great now. I think I prefer this solution over #85.

hagenw:
OK, I finished #86 and requested a review there.

hagenw:
Closing as we have decided to go for the solution presented in #86.

PR #86, opened 2022-12-15
DOC: create single class/function pages with autosummary

hagenw:
Alternative implementation to https://github.com/audeering/audinterface/pull/85.

This introduces a new structure for our documentation, that I would like to propose for all our packages.
To avoid very long pages and provide also links for class methods and attributes in the TOC, it uses [sphinx.ext.autosummary](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html) to automatically split up the documentation into sub-pages.
It will generate one sub-page for each function or class.

## autosummary and templates

`sphinx.ext.autosummary` is able to create TOC entries and corresponding sub-pages when used with the :toctree: option. For all (sub)-modules we now create files inside `docs/api-src` that have such an entry (see sub-section below). 

During building the docs, the src files are copied to `docs/api` and `sphinx.ext.autosummary` generates additional, so called stub files, for each class and function inside `docs/api`.

What exactly is included in pages that are automatically generated can then be configured by templates. The important ones are `docs/_templates/autosummary/class.rst` and `docs/_templates/autosummary/function.rst` in which we define how a sub-page for a class or function should be presented and also how it is named inside the TOC. 

The definition of the templates could also be done in the theme, but it might be that we want to adapt them inside projects, e.g. here we want to list the special method `__call__`, but in other projects we might want to not list it.
So for now I would propose we include the templates in the single projects.

All changes are explained in detail in the following sub-sections.

## Predictable URL structure

The URL of a given class/function/method etc. is completely given by the module name and all the other names, e.g.

| Name | URL |
| --- | --- |
| `audinterface` | `base-url/api/audinterface.html` |
| `audinterface.Feature` | `base-url/api/audinterface.Feature.html` |
| `audinterface.Feature.process_index()` | `base-url/api/audinterface.Feature.html#process_index` |
| `audinterface.Feature.name` | `base-url/api/audinterface.Feature.html#name` |
| `audinterface.utils` | `base-url/api/audinterface.utils.html` |
| `audinterface.utils.read_audio()` | `base-url/api/audinterface.utils.read_audio.html` |

The downside is that this will break existing links to the API in other published documentations.
For that reason it would be my goal to update all our API documentations with this new schema until mid of January, then we should be able to eliminate all broken links.

## Inclusion in docs via (sub)-module files

As before we have to create entry files for each (sub)-module, but those now have an easier structure,
e.g. `docs/api/audinterface.utils.rst`:

```rst
audinterface.utils
==================

.. automodule:: audinterface.utils

.. autosummary::
    :toctree:
    :nosignatures:

    read_audio
    signal_index
    sliding_window
    to_timedelta
```

Which will result in the following rendered page:

image

showing the following TOC entry

image

## Listing of all class methods/properties/attributes

We now get single pages for each class, that still lists all methods and attributes on the same page to make searching for keywords easier. All methods and attributes are listed in the TOC, whereas methods are marked with `()` at the end, e.g.

image

Corresponding class page. The biggest difference to the current presentation is that each method/attribute gets an extra heading to generate the TOC entry and that each method/attribute is listed with class name in front of it and a red left border:

image

## Updates to Example section

* Rename to "Examples"


frankenjoe:
What's the correct building command? With `python -m sphinx -W docs/ build/sphinx/html -b html` I'm getting:

```
Warning, treated as error:
audinterface/core/feature.py:docstring of audinterface.core.feature.Feature.__call__:1:duplicate object description of audinterface.Feature.__call__, other instance in api/audinterface.Feature, use :noindex: for one of them
```

audinterface/core/utils.py
```diff
@@ -221,28 +221,34 @@ def signal_index(
     Raises:
         ValueError: if ``start`` and ``ends`` differ in size
 
-    Example:
+    Examples:
+
         >>> signal_index(0, 1.1)
         MultiIndex([('0 days', '0 days 00:00:01.100000')],
                    names=['start', 'end'])
+
```
frankenjoe, line 229: Why the blank line here?

hagenw: To split independent examples into different blocks:

image

frankenjoe:
I think we should intend "Apply interface ..."

image

audinterface/core/utils.py
frankenjoe: Mhh, I'm not convinced :)

docs/api-src/audinterface.rst
```diff
@@ -0,0 +1,36 @@
...
+You can inherit from the classes
+or just instantiate them
+to get some standard implementations of those methods.
+The following classes are provided:
+
+.. autosummary::
+    :toctree:
+    :nosignatures:
+
+    Feature
```
frankenjoe, line 33: Cool, looks much easier now :)

hagenw: Yes, I also found it nice that we can now simply list all the function/classes here.

The only thing that might be slightly confusing is that we have the manual written files under `docs/api-src` and the automatically created files under `docs/api`. But for debugging it is easier to have separate folders and if something does not work you can also just delete `docs/api` and try again.

hagenw:
> I think we should indent "Apply interface ..."
> 
> image

This is something we need to handle inside the theme. At least the theme is repsonsible that the code block is indented. Something that is not the common behavior, e.g. compare https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.abs.html where the code block is not indented:

image

So we have two options:
* not indent the code block
* indent the text in the examples section

I'm not sure yet what is the better solution, but I would not make it part of this pull request, but change it inside `sphinx-audeering-theme` instead.

audinterface/core/utils.py
hagenw: Would you then stick to the compact format all the time, or split it up when we write explanatory text inbetween like we do here:

image

frankenjoe: yes, exactly

frankenjoe: I mean yes, only break the compact form when we write explanatory text :)

docs/api-src/audinterface.rst
frankenjoe: Yes, we should probably document this.

frankenjoe:
> I'm not sure yet what is the better solution, but I would not make it part of this pull request, but change it inside `sphinx-audeering-theme` instead.

I would intend it both, the code block and the explanatory text, otherwise it looks like it no longer belongs to the `Examples` section.

hagenw:
> > I'm not sure yet what is the better solution, but I would not make it part of this pull request, but change it inside `sphinx-audeering-theme` instead.
> 
> I would intend it both, the code block and the explanatory text, otherwise it looks like it no longer belongs to the `Examples` section.

Turns out that this is not trivial (and maybe the reason why no other documentation does it).

The problem is that the "Example(s)" section of the docstring and others like "Notes" are converted to a simple rubric entry and its content is not put inside a `<div>`. For the code samples we still have CSS class names corresponding to the code highlight, which we currently use to indent the code examples:

```css
.rst-content dl.class div.highlight,
.rst-content dl.function div.highlight,
.rst-content dl.method div.highlight {
    margin-left: 24px;
}
```

but as the following screenshot shows, the text is only put in a simple `<p>` statement without any class name that we could use to identify and style it:

image

In the short term I see only two possible solutions:

1. don't break up the examples into different peaces and leave the explanatory text inside the code blocks as comments
2. break up the examples when we have explanatory text and change `sphinx-audeering-theme` to not indent the examples

A result of the later option would look like this:

image

---

So maybe we don't split up the examples yet (solution 1.) until we find a solution how to indent the whole examples section?

frankenjoe:
> So maybe we don't split up the examples yet (solution 1.) until we find a solution how to indent the whole examples section?

Yes, I would go with this solution. This also has the advantage that we don't have to touch the examples in our packages, except for maybe changing 'Example' to 'Examples' :)

hagenw:
I decided now also for the option to not split up the examples and changed the code and description accordingly.

docs/api-src/audinterface.rst
hagenw: Should I add a section to `CONTRIBUTING.rst` or should we target to document it at our central place at http://devops.pp.audeering.com/sphinx/?

audinterface/core/utils.py
hagenw: As discussed below, we do no longer split up the examples.

docs/api-src/audinterface.rst
frankenjoe: Is http://devops.pp.audeering.com/sphinx/ available outside audeering?

docs/conf.py
```diff
@@ -1,6 +1,10 @@
+import os
+import shutil
 import subprocess
 from datetime import date
```
frankenjoe, line 4:
```suggestion
from datetime import date
import os
import shutil
import subprocess
```

audinterface/core/feature.py
```diff
@@ -519,7 +523,7 @@ def process_index(
             root: str = None,
             cache_root: str = None,
     ) -> pd.DataFrame:
-        r"""Extract features from an index conform to audformat_.
```
frankenjoe, line 522: Why did you remove the link?

docs/api-src/audinterface.rst
hagenw: No, but I think we also have two different use cases of documenting this:

* write documentation on how to setup a new Python package including documentation. For this http://devops.pp.audeering.com/sphinx/ together with the `audfoo` package seems the right place to me
* then in `CONTRBUTING.rst` we could document what you have todo when you add a new function or class, e.g. where to add it in the documentation. But so far, we also didn't do it.

audinterface/core/feature.py
hagenw: I did remove it as in the summary table that `autosummary` creates (which we no longer use here, but which was used in #85), the link does not work. I think in general it is a good idea if the first line of the docstring works without extra input. But I'm also fine with adding the link back.

docs/api-src/audinterface.rst
hagenw: So for now I would propose that we stay with the current version of `CONTRIBUTING.rst` and during the upcoming weeks first update the general sphinx documentation and the `audfoo` package. If we then still see the need for adjusting `CONTRIBUTING.rst` in individual packages we can do this as well.

audinterface/core/feature.py
frankenjoe: I would leave it as it might not be obvious for a user what `audformat` is.

docs/api-src/audinterface.rst
frankenjoe: ok

audinterface/core/feature.py
hagenw: I re-enabled the links in all docstrings.

PR #89, opened 2022-12-23
Add support for Python 3.10

hagenw:
Add test for Python 3.10 under Ubuntu and mention Python 3.10 as supported version in `setup.cfg`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest, windows-latest, macos-latest ]
-        python-version: [ 3.8, 3.9 ]
+        python-version: [ '3.8', '3.9', '3.10' ]
 
     steps:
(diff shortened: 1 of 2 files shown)


hagenw:
In https://github.com/audeering/audinterface/actions/runs/3764772386/jobs/6399549178
we got the following error when using `audb` in the docs:

```
  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/artifactory.py", line 1413, in stat
    return self._accessor.stat(pathobj=pathobj)
TypeError: stat() missing required argument 'path' (pos 1)
```

Now, I restarted the job by rebasing on `main` and the error is gone.
First I thought `dohq-artifactory` was the problem, but it installed `0.7.742` in both cases.

Also the Python version is identical to Python 3.10.9, so it could also not be that something underlying in `Pathlib` was fixed.

As the error was happening anyway only when building the docs, I think we can ignore it and don't have to figure out which external package + version caused it.

PR #96, opened 2023-02-02
Add preserve_index argument to process_index()

### LINKED ISSUE #71: Support selection of returned index type in process_index()
At the moment the `process_index()` method is always returning a segmented index which works great in most circumstances, but it can also have some undesired consequences:

```
import audb
import audinterface
import numpy as np


def rms_db(signal, sampling_rate):
    return 20 * np.log10(np.sqrt(np.mean(np.square(signal))))


interface = audinterface.Process(process_func=rms_db)
db = audb.load('emodb', version='1.3.0', sampling_rate=16000, mixdown=True, format='wav')
df = db['emotion'].get()
df['level'] = interface.process_index(df.index)
```
works without throwing an error, but if you inspect the results you will see that they are empty as the indices of `df` and the returned series do not match:
```python
>>> df
                                                     emotion  emotion.confidence  level
file                                                                                   
/home/audeering.local/hwierstorf/audb/emodb/1....  happiness                 0.9    NaN
/home/audeering.local/hwierstorf/audb/emodb/1....    neutral                 1.0    NaN
/home/audeering.local/hwierstorf/audb/emodb/1....      anger                 1.0    NaN
```

What you need todo instead is:
```python
>>> df['level'] = interface.process_index(df.index).values
>>> df
                                                     emotion  emotion.confidence      level
file                                                                                       
/home/audeering.local/hwierstorf/audb/emodb/1....  happiness                 0.9 -21.690142
/home/audeering.local/hwierstorf/audb/emodb/1....    neutral                 1.0 -18.040704
/home/audeering.local/hwierstorf/audb/emodb/1....      anger                 1.0 -20.394533
```

This could be mitigated if we add an option to select to return the original index type instead. On the other hand the user will most likely also forget to set the argument and could use `.values` instead as shown here.

So I'm not decided if we need todo something here, but I thought it is worse reporting this problem.

hagenw:
Closes #71 

This adds the argument `preserve_index` to `Process.process_index()` and `Feature.process_index()`. If `True` it will return a filewise index instead of a segmented index if provided with a filewise index. This has the advantage that one can directly assign the result to a dataframe with the original index as outlined in #71.

It's not implemented in the most elegant way as nearly all internal private functions and the caching requires segmented index. If `preserve_index=True` the index is first converted to a segmented index anyway and also the results are always cached as a segmented index. We just convert the index back to a segmented index before returning it to the user.

### Docstring and example for Process

image

image

### Docstring and example for Feature

image

image

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -560,7 +560,6 @@ class Feature:
 
         """
         cache_path = None
-        index_type = audformat.index_type(index)
 
         if cache_root is not None:
             cache_root = audeer.mkdir(cache_root)
@@ -582,11 +581,7 @@ class Feature:
         if self.process.segment is None and preserve_index:
             # Convert segmented index to filewise index
             # if original index was filewise
-            if index_type == audformat.define.IndexType.FILEWISE:
-                files = df.index.get_level_values(
-                    audformat.define.IndexField.FILE
-                )
-                df.index = audformat.filewise_index(files)
+            df.index = index
 
         return df
 
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -204,6 +204,10 @@ class Feature:
                                                            mean       std
         file            start  end
         wav/03a01Fa.wav 0 days 0 days 00:00:01.898250 -0.000311  0.082317
+        >>> interface.process_index(index, root=db.root, preserve_index=True)
+                        mean       std
+        file
+        wav/03a01Fa.wav -0.000311  0.082317
         >>> # Apply interface with a sliding window
         >>> interface = Feature(
         ...     ['mean', 'std'],
@@ -520,6 +524,7 @@ class Feature:
             self,
             index: pd.Index,
             *,
+            preserve_index: bool = False,
             root: str = None,
             cache_root: str = None,
     ) -> pd.DataFrame:
@@ -537,6 +542,12 @@ class Feature:
 
         Args:
             index: index with segment information
+            preserve_index: if ``True``
+                and :attr:`audinterface.Feature.process.segment` is ``None``
+                the returned index
+                will be of same type
+                as the original one,
+                otherwise always a segmented index is returned
             root: root folder to expand relative file paths
             cache_root: cache folder (see description)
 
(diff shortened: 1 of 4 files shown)


audinterface/core/feature.py
```diff
@@ -567,6 +579,15 @@ def process_index(
             if cache_path is not None:
                 df.to_pickle(cache_path, protocol=4)
 
+        if self.process.segment is None and preserve_index:
+            # Convert segmented index to filewise index
+            # if original index was filewise
+            if index_type == audformat.define.IndexType.FILEWISE:
```
frankenjoe, line 585: Can't we simply do: `df.index = index`?

hagenw: You are right.
I updated the code here and in `process.py`.

frankenjoe: Thanks. Since you are no longer checking if the input is a filewise index, this will also preserve `NaT` values for segmented indices, which I think is what a user would expect. But as far as I see this case is not covered by the tests, yet.

hagenw: I added the `NaT` to the tests and restructured them to provide `preserve_index` as a pytest parameter.

tests/test_process.py
```diff
@@ -604,29 +599,62 @@ def test_process_index(tmpdir, num_workers, multiprocessing):
...
-        signal, sampling_rate = audinterface.utils.read_audio(
-            file, start=start, end=end, root=root
-        )
-        np.testing.assert_equal(signal, value)
+    y = process.process_index(
+        index,
+        preserve_index=preserve_index,
+    )
+    if preserve_index:
+        for path, value in y.items():
```
frankenjoe, line 624: Maybe add `pd.testing.assert_frame_equal(y.index, index)` here and in the other places?

frankenjoe: Sorry, I mean `assert_index_equal()` of course :)

hagenw: I added it.

hagenw:
Tests on Ubuntu will fail until Ubuntu or Github have solved https://github.com/actions/runner-images/issues/7048

PR #97, opened 2023-02-02
Use audmath in audinterface.utils.to_timedelta()

hagenw:
This updates `audinterface.utils.to_timedelta()` to use `audmath.duration_in_seconds()` as single source of truth for conversion to seconds.

The behavior of `audinterface.utils.to_timedelta()` is not changed as we have only a single `sampling_rate` argument for a whole list of duration values:
* float and integer values will always be tyreated as representing duration in seconds
* only string values without a unit are expected to represent samples, and we raise an error if such a value is given, but no `sampling_rate` value

image

### DIFF
--- a/audinterface/core/process.py
+++ b/audinterface/core/process.py
@@ -256,8 +256,10 @@ class Process:
             end: pd.Timedelta = None,
     ) -> pd.Series:
 
-        start = utils.to_timedelta(start, self.sampling_rate)
-        end = utils.to_timedelta(end, self.sampling_rate)
+        if start is not None:
+            start = utils.to_timedelta(start, self.sampling_rate)
+        if end is not None:
+            end = utils.to_timedelta(end, self.sampling_rate)
 
         signal, sampling_rate = utils.read_audio(
             file,
@@ -687,8 +689,10 @@ class Process:
                 index,
             )
         else:
-            start = utils.to_timedelta(start, sampling_rate)
-            end = utils.to_timedelta(end, sampling_rate)
+            if start is not None:
+                start = utils.to_timedelta(start, sampling_rate)
+            if end is not None:
+                end = utils.to_timedelta(end, sampling_rate)
             return self._process_signal(
                 signal,
                 sampling_rate,
--- a/audinterface/core/utils.py
+++ b/audinterface/core/utils.py
@@ -407,7 +407,7 @@ def to_timedelta(
     The single duration values
     support all formats
     mentioned in :func:`audmath.duration_in_seconds`,
-    like ``'2 ms'``, or ``pd.to_timedelta(2, 's')``.
+    like ``'2 ms'``, or ``pandas.to_timedelta(2, 's')``.
     The exception is
     that float and integer values
     are always interpreted as seconds
--- a/audinterface/core/utils.py
+++ b/audinterface/core/utils.py
@@ -404,20 +404,15 @@ def to_timedelta(
 ) -> typing.Union[pd.Timedelta, typing.List[pd.Timedelta]]:
     r"""Convert duration value(s) to :class:`pandas.Timedelta`.
 
-    If duration is given as string without unit,
-    it is treated as samples
-    and requires that ``'sampling_rate'`` is not ``None``.
-
-    .. warning::
-
-        From version 1.0.0 not providing ``sampling_rate``
-        no longer raises an error
-        when duration is provided
-        as a string value without unit,
-        but the string value will be interpreted as seconds;
-        if ``sampling_rate`` is provided
-        integer and float duration values
-        will be interpreted as seconds.
+    The single duration values
+    support all formats
+    mentioned in :func:`audmath.duration_in_seconds`,
+    like ``'2 ms'``, or ``pd.to_timedelta(2, 's')``.
+    The exception is
+    that float and integer values
+    are always interpreted as seconds
+    and strings without unit
+    always as samples.
 
     Args:
         durations: duration value(s).
@@ -452,12 +447,7 @@ def to_timedelta(
 
     """  # noqa: E501
     def duration_in_seconds(duration, sampling_rate):
-        """Helper function to convert to seconds.
-
-        We will remove this helper function in version 1.0.0 of audinterface
-        and replace it by calling directly audmath.duration_in_seconds().
-
-        """
+        """Helper function to convert to seconds."""
         if not isinstance(duration, str):
             # force non-string values to represent seconds
             sampling_rate = None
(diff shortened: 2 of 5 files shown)


frankenjoe:
> We need a little bit of extra code to force that `'200'` can only represent samples, and to raise an error if you don't provide a sampling rate in this case.

Is your motivation that code that broke before still breaks or is there another reason to not interpret `'200'`  as seconds if sampling rate is not given?

hagenw:
> > We need a little bit of extra code to force that `'200'` can only represent samples, and to raise an error if you don't provide a sampling rate in this case.
> 
> Is your motivation that code that broke before still breaks [...]?

Exactly.

frankenjoe:
It's a bit unfortunate that we now have different behavior between `audinterface.utils.to_timedelta()` and `audmath.duration_in_seconds()`, so I wonder if we should simply deprecate  `audinterface.utils.to_timedelta()` instead?

hagenw:
> It's a bit unfortunate that we now have different behavior between `audinterface.utils.to_timedelta()` and `audmath.duration_in_seconds()`, so I wonder if we should simply deprecate `audinterface.utils.to_timedelta()` instead?

I would not deprecate it as it also converts to a timedelta object.
Or do you mean deprecate it's current behavior and change it to be identical to `audmath.duration_in_seconds()` in a future version?

We have a similar problem in https://github.com/audeering/audiofile/pull/109. So we might also think about changing the behavior of `audmath.duration_in_seconds()` to not support `'200'` to represent samples. But I think I would be more in favor of changing `audinterface.utils.to_timedelta()`.

frankenjoe:
> I would not deprecate it as it also converts to a timedelta object.

Sure, but now a user can simply do `pd.Timedelta(audmath.duration_in_seconds(dur), 's')`.

If we decide to keep `audinterface.utils.to_timedelta()`, I would suggest we change the behavior, even if it means a breaking change.

frankenjoe:
Another feature of `audinterface.utils.to_timedelta()` that it also works on lists, which is indeed an argument to keep it, but adapt its behavior to `audmath.duration_in_seconds()`.

hagenw:
There are two more differences compared to `audmath.duration_in_seconds()`:

* `audinterface.utils.to_timedelta()` also supports `NaT`, which is not supported in 
`audmath.duration_in_seconds()`. Maybe we should check if we can support it there without requiring `pandas`.
What works in `audmath.duration_in_seconds()` as well is `None`:
```python
>>> audmath.duration_in_seconds(None)
nan
```
* `audinterface.utils.to_timedelta()` always interprets `float` and `int` values as seconds independent if `sampling_rate` is provided or not

I'm afraid we made ` audmath.duration_in_seconds()` complicated to use as the behavior depends not only on the duration value, but also on `sampling_rate`, but for some applications like `audiofile.read()` there is no way that a user can set the sampling rate as it is always provided by the file. So maybe it would be better to adjust the behavior of `audmtah.duration_in_seconds()` to use only `duration` to determine what should happen for the conversion?

frankenjoe:
> But for some applications like audiofile.read() there is no way that a user can set the sampling rate as it is always provided by the file

To me it seems fine, as long as we document that inside `audiofile.read()` we always call `audmath.duration_in_seconds()` with the sampling rate of the file. Same with `audinterface.Feature`.

But what could be confusing if `audinterface.utils.to_timedelta()` and `audmath.duration_in_seconds()` do not work in the same way, where it is up to the user to provide a sampling rate or not.

hagenw:
> There are two more differences compared to `audmath.duration_in_seconds()`:
> 
> * `audinterface.utils.to_timedelta()` also supports `NaT`, which is not supported in
>   `audmath.duration_in_seconds()`. Maybe we should check if we can support it there without requiring `pandas`.
>   What works in `audmath.duration_in_seconds()` as well is `None`:
> 
> ```python
> >>> audmath.duration_in_seconds(None)
> nan
> ```

With [audmath 1.2.1](https://github.com/audeering/audmath/pull/35) we now also support `NaT` and don't need to add extra code to `to_timedelta()`. I just needed to add some extra `if start is not None` to `audinterface/core/process.py` to ensure `start` and `end` values containing `None` are not replaced by `nan`.
 
> * `audinterface.utils.to_timedelta()` always interprets `float` and `int` values as seconds independent if `sampling_rate` is provided or not

I think changing this behavior shouldn't be problematic as in cases where we provided `'200'` as samples the sampling rate has been always provided before (otherwise an error would have occurred).

The question is how we should handle the deprecation in `audinterface`. One way would be to extend the error message that is raised when `sampling_rate` is not provided for a string value, e.g.

```
                   "You have to provide 'sampling_rate' "
                    "when specifying the duration in samples "
                    f"as you did with '{duration}'. "
                    "NOTE: this will no longer raise an error "
                    "in version 1.0.0, "
                    f"but interpret '{duration}' in seconds."
```

and state in the docstring under the raises entry that it will be removed in 1.0.0:

image

frankenjoe:
> I think changing this behavior shouldn't be problematic as in cases where we provided `'200'` as samples the sampling rate has been always provided before (otherwise an error would have occurred).

Exactly, so I would argue there is no need for a deprecation warning at all.

hagenw:
> > I think changing this behavior shouldn't be problematic as in cases where we provided `'200'` as samples the sampling rate has been always provided before (otherwise an error would have occurred).
> 
> Exactly, so I would argue there is no need for a deprecation warning at all.

You are right, I will remove that.

But we have a much bigger problem with the other case: integers and floats are interpreted as samples if `sampling_rate` is given. But we are only provide a single sampling rate for the whole list, e.g. at the moment we have:

```python
>>> to_timedelta([1, '2000'], 1000)
[Timedelta('0 days 00:00:01'), Timedelta('0 days 00:00:02')]
```

This behavior will no longer be possible with the current behavior of `audmath.duration_in_seconds()`.

hagenw:
I see two solutions:

* we change `audmath.duration_in_seconds()` to always treat float or integer values as duration in seconds
* we don't adjust `audinterface.utils.to_timedelta()` to behave the same as `audmath.duration_in_seconds()`, but state that float and integer values are always treated as duration in seconds.

I would be more in favor of the second part. And if we do this, we might also stay with allowing a string without unit to only represent samples.

frankenjoe:
Too bad, but it indeed seems we cannot align those two functions. Since I still think that our current implementation of `audmath.duration_in_seconds()` makes a lot sense since we do not deal with list input, I would also opt for your second solution.

hagenw:
We have to mention the `ValueError` that is raised if the string has not a proper `<value><unit>` format in the docstring.
I will add it.

hagenw:
I added the other `ValueError` now to the docstring:

image

and also added two tests test for it.

PR #99, opened 2023-02-07
CI: cache emodb database

hagenw:
As we use `emodb` loaded by `audb` in the documentation examples, it will make sense to cache the data as we do in https://github.com/audeering/audb to avoid reloading the data with every pipeline from the external server.


hagenw:
Cache is working:

image

PR #101, opened 2023-02-09
Add support for Python 3.11

hagenw:
Add test for Python 3.11 under Ubuntu and mention Python 3.11 as supported version in `setup.cfg`.


hagenw:
For Python 3.11 support we need to wait until `numba` does support it, see https://github.com/numba/numba/issues/8304

hagenw:
`numba` 0.57.0 was [released](https://github.com/numba/numba/releases/tag/0.57.0) and now we can also support Python 3.11 here.

Issue #100, opened 2023-02-08
Speed up feature extraction


frankenjoe:
When extracting features with `Feature` we currently rely on `Process` under the hood, which returns a `pd.Series` with feature vectors. We then convert these to a list and afterwards call `pd.concat(list)` to combine them to a single matrix. The last step can take quite long (sometimes as long or longer as the feature extraction itself). We could speed this up if we pre-allocate a matrix beforehand and directly assign the values. At least when not processing with a sliding window this should be possible.

frankenjoe:
To demonstrate there's quite some room for improvement:

```python
import pandas as pd

import audb
import audinterface
import audiofile


db = audb.load(
    'emodb',
    version='1.3.0',
    format='wav',
    sampling_rate=16000,
    mixdown=True,
)
files = db.files

def process_func(x, sr):
    return [x.mean(), x.std()]

# slow

feature = audinterface.Feature(
    ['mean', 'std'],
    process_func=process_func,
)

t = time.time()
df = feature.process_files(files)
print(time.time() - t)

# fast

t = time.time()
data = np.empty(
    (len(files), 2),
    dtype=np.float32,
)

for idx, file in enumerate(files):
    signal, sampling_rate = audiofile.read(file)
    data[idx, :] = process_func(
        signal,
        sampling_rate,
    )

df_fast = pd.DataFrame(
    data,
    index=df.index,
    columns=df.columns,
)
print(time.time() - t)

pd.testing.assert_frame_equal(df, df_fast)
```
```
5.972992181777954
0.17418813705444336
```

hagenw:
> We then convert these to a list 

I guess the idea for a solution is to avoid this step?

frankenjoe:
Yes, especially the concatenation of the DataFrames seems awefully slow. So the idea would be to create a matrix of the expected size (samples x features) and directly assign the extracted features. This is of course only possible if no sliding window is selected as otherwise we cannot know the shape of the final matrix.

hagenw:
After #102, #103, and #104 the above test now returns the following for me:

```
0.23550820350646973                                                                                 
0.17041683197021484
```

Can we close here, or is there further room for improvement?

hagenw:
I created https://github.com/audeering/audinterface/issues/106 to track `Segment` and will close here.

PR #102, opened 2023-03-09
Speed up Predict.process_*()

frankenjoe:
Relates to #100 

Speeds up `Predict.process_*()` by avoiding calls to `pd.concat()` on a lot of small `pd.Series` objects. Instead, we now concatenate lists and create the final `pd.Series` from those.

### Benchmark

```python
import time
import audb
import audinterface


db = audb.load(
    'emodb',
    version='1.3.0',
    format='wav',
    sampling_rate=16000,
    mixdown=True,
)

def process_func(x, sr):
    return [x.mean(), x.std()]

process = audinterface.Process(process_func=process_func)

t = time.time()
process.process_files(db.files)
print(time.time() - t)
```

* main branch: ~4.5 s
* this branch: ~0.3 s

### DIFF
--- a/audinterface/core/process.py
+++ b/audinterface/core/process.py
@@ -1,5 +1,6 @@
 import errno
 import inspect
+import itertools
 import os
 import typing
 import warnings
@@ -254,7 +255,12 @@ class Process:
             root: str = None,
             start: pd.Timedelta = None,
             end: pd.Timedelta = None,
-    ) -> pd.Series:
+    ) -> typing.Tuple[
+        typing.List[typing.Any],
+        typing.List[str],
+        typing.List[pd.Timedelta],
+        typing.List[pd.Timedelta],
+    ]:
 
         if start is not None:
             start = utils.to_timedelta(start, self.sampling_rate)
@@ -267,7 +273,8 @@ class Process:
             end=end,
             root=root,
         )
-        y = self._process_signal(
+
+        y, files, starts, ends = self._process_signal(
             signal,
             sampling_rate,
             idx=idx,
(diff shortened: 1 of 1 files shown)


audinterface/core/process.py
```diff
@@ -411,7 +415,16 @@ def process_files(
             task_description=f'Process {len(files)} files',
         )
         self.verbose = verbose
-        return pd.concat(y)
+
+        y = list(itertools.chain.from_iterable([x[0] for x in xs]))
```
hagenw, line 419: Why is 

```python
y = [x[0] for x in xs]
```
not working?

frankenjoe: We have a list of lists that we want to flatten.

hagenw: Ah, ok, so `x` is not a tuple containing single `(y, file, start, end)` entries, but `(y, files, starts, ends)` lists, correct?

frankenjoe: yes

hagenw:
We have now code that we repeat at some places. In principle we could replace it by the following function:

```python
def _create_series_from_list(xs, skip_file_level=False):

    y = list(itertools.chain.from_iterable([x[0] for x in xs]))
    starts = list(itertools.chain.from_iterable([x[2] for x in xs]))
    ends = list(itertools.chain.from_iterable([x[3] for x in xs]))

    if skip_file_level:
        index = utils.signal_index(starts, ends)
    else:
        files = list(itertools.chain.from_iterable([x[1] for x in xs]))

    return pd.Series(y, index)
```

But I'm not completely sure if this might it harder to understand what is going on inside specific methods, e.g.

```python
        xs = audeer.run_tasks(
            self._process_file,
            params,
            num_workers=self.num_workers,
            multiprocessing=self.multiprocessing,
            progress_bar=self.verbose,
            task_description=f'Process {len(index)} segments',
        )
        y = _create_series_from_list(xs)
```

Did you already thought about this?

frankenjoe:
Yes, I thought about it, but as you say, it makes the code more complicated to follow and it's very little code we duplicate. So I think I would leave it as is.

PR #107, opened 2023-03-13
Speed up Segment.process_*()

### LINKED ISSUE #106: Speed up audinterface.Segment
Similar to #100 we want to speed up `audinterface.Segment` when processing several files.

At the moment we get:

```python
import pandas as pd

import audb
import audinterface
import audiofile


db = audb.load(
    'emodb',
    version='1.3.0',
    format='wav',
    sampling_rate=16000,
    mixdown=True,
)
files = db.files

def process_func(x, sr):
    return pd.MultiIndex.from_tuples(
        [(pd.to_timedelta(0), pd.to_timedelta(1))],
        names=['start', 'end'],
    )

# slow

segment = audinterface.Segment(
    process_func=process_func,
)

t = time.time()
index_slow = segment.process_files(files)
print(time.time() - t)

# fast

t = time.time()
starts = []
ends = []
for idx, file in enumerate(files):
    signal, sampling_rate = audiofile.read(file)
    index = process_func(
        signal,
        sampling_rate,
    )
    starts += list(index.get_level_values('start'))
    ends += list(index.get_level_values('end'))

index_fast = pd.MultiIndex.from_arrays(
    [files, starts, ends],
    names=['file', 'start', 'end'],
)
print(time.time() - t)

pd.testing.assert_index_equal(index_slow, index_fast)
```
```
4.128075122833252                                                                                   
1.7938168048858643
```

frankenjoe:
Closes #106 

Avoids calling `audformat.utils.union()` by collecting files, starts and ends in a list and use them to directly create the final index.

### DIFF
--- a/audinterface/core/segment.py
+++ b/audinterface/core/segment.py
@@ -325,19 +325,19 @@ class Segment:
         .. _audformat: https://audeering.github.io/audformat/data-format.html
 
         """
-        series = self.process.process_files(
+        y = self.process.process_files(
             files,
             starts=starts,
             ends=ends,
             root=root,
         )
-        if len(series) == 0:
+        if len(y) == 0:
             return audformat.filewise_index()
 
         files = []
         starts = []
         ends = []
-        for idx, ((file, start, _), index) in enumerate(series.items()):
+        for idx, ((file, start, _), index) in enumerate(y.items()):
             files.extend([file] * len(index))
             starts.extend(index.levels[0] + start)
             ends.extend(index.levels[1] + start)
--- a/audinterface/core/segment.py
+++ b/audinterface/core/segment.py
@@ -333,16 +333,16 @@ class Segment:
         )
         if len(series) == 0:
             return audformat.filewise_index()
-        objs = []
+
+        files = []
+        starts = []
+        ends = []
         for idx, ((file, start, _), index) in enumerate(series.items()):
-            objs.append(
-                audformat.segmented_index(
-                    files=[file] * len(index),
-                    starts=index.levels[0] + start,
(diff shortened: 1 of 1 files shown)


frankenjoe:
When I run the example from #106 I get:

* main branch: 3.74
* this branch: 0.89

frankenjoe:
I wonder if we can optimize `aufdformat.utils.union()` in a similar way...

audinterface/core/segment.py
```diff
@@ -533,7 +535,21 @@ def process_signal_from_index(
             progress_bar=self.process.verbose,
             task_description=f'Process {len(index)} segments',
         )
-        index = audformat.utils.union(y)
```
frankenjoe, line 536: I was surprised to see that we were using `audformat.utils.union()` here also for the case with a file level. I was not expecting this could work...

frankenjoe: Ah right, forgot we changed it to support all indices. This will make it difficult to optimize in a similar way as we do here.

audinterface/core/segment.py
```diff
@@ -325,24 +325,24 @@ def process_files(
...
+
+        files = []
+        starts = []
+        ends = []
+        for (file, start, _), index in y.items():
+            files.extend([file] * len(index))
+            starts.extend(index.levels[0] + start)
+            ends.extend(index.levels[1] + start)
+
+        return audformat.segmented_index(files, starts, ends)
```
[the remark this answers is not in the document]
hagenw: It might be that we want to handle this case with some custom error message, but at the moment I would say we just assume that the segment object does not return identical segments.

PR #110, opened 2023-03-28
Fix: do not truncate string features

### LINKED ISSUE #109: String output in processing function is truncated in audinterface version 1.0.0
When running the following code, I get incorrect results for version 1.0.0:

```python
import random
import audb
import audinterface

db = audb.load(
    'emodb',
    version='1.3.0',
    verbose=False,
)
df = db['files'].get()

test_labels = ['anger', 'happiness', 'neutral', 'sadness']
def process_func(x, fs):
    return random.sample(test_labels, 1)

    'emotion',
    process_func=process_func
)
random.seed(2)
prediction = interface.process_index(df.index)
```

For version 1.0.0, all labels are truncated to the same length as the first randomly selected label.
For instance, if the first randomly chosen label is `anger`, then `neutral` is cut to `neutr`, `happiness` to `happi`, and `sadness` to `sadne`.

For audinterface version 0.10.2, the code works as expected.

frankenjoe:
Closes #109

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -767,8 +767,6 @@ class Feature:
                 dtype=object,
             )
 
-        num = len(y)
-
         if (
                 self.win_dur is not None and
                 self.process_func_applies_sliding_window
@@ -828,8 +826,6 @@ class Feature:
         else:
 
             index = y.index
-            dtype = self._values_to_frame(y[0]).dtype
-            shape = (num, len(self.column_names))
             data = [self._values_to_frame(values) for values in y]
 
         data = np.concatenate(data)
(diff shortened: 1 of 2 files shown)


frankenjoe:
I was already afraid of running into problems when pre-allocating the numpy array. I should also add some tests to cover it in the future.

frankenjoe:
I was already afraid of running into problems when pre-allocating the numpy array. I should also add some tests to cover it in the future.

frankenjoe:
> I should also add some tests to cover it in the future.

done

hagenw:
The new test fails indeed for the `main` branch.

I also tested performance of your new changes. The execution time for this branch and the main branch is the same when executing the benchmark proposed in https://github.com/audeering/audinterface/pull/103

hagenw:
The title of this pull request is slightly too long, maybe rename to:
"Fix: do not truncate string features"

frankenjoe:
> The title of this pull request is slightly too long, maybe rename to:
> "Fix: do not truncate string features"

done

PR #124, opened 2023-07-03
Fix rounding of start/end in utils.read_signal()

### LINKED ISSUE #123: process_signal() and process_file() can return different results
First we get some data and define an interface:

```python
import audb
import audinterface
import audiofile

files = audb.load_media('emodb', 'wav/13b09La.wav', version='1.4.1')
signal, sampling_rate = audiofile.read(files[0])
interface = audinterface.Process(process_func=lambda x, sr: x.sum())
```

Then we select a part of the audio for which we get the same result:
```python
import pandas as pd

start = pd.Timedelta('0 days 00:00:00.120000')
end = pd.Timedelta('0 days 00:00:02.440000')
```
then we get
```python
>>> interface.process_file(files[0], start=start, end=end).tolist()
[-21.31195068359375]

>>> interface.process_signal(signal, sampling_rate, start=start, end=end).tolist()
[-21.31195068359375]
```

Then we use end and start values from https://github.com/audeering/audinterface/issues/122 for which we got different results there:

```python
start = pd.Timedelta('0 days 00:00:01.140000')
end = pd.Timedelta('0 days 00:00:01.560000')
```

then we get

```python
>>> interface.process_file(files[0], start=start, end=end).tolist()
[1.95721435546875]

>>>  interface.process_signal(signal, sampling_rate, start=start, end=end).tolist()
[2.218841552734375]
```

hagenw:
Closes #123 

For certain values of `start` and `end` `process_file(..., start=start, end=end)` returned a different result than `process_signal(..., start=start, end=end)` as the duration values were rounded differently inside `audinterface.utils.read_audio()`, which is used in `process_file()`.

The behavior is now fixed inside ` audinterface.utils.read_signal()` and `process_file()` returns now the same results as `process_signal()`.

The fix uses `audmath.samples()` as a ground truth for converting to samples and relies on the fix of `audiofile` to also use `audmath.samples()` as ground truth (https://github.com/audeering/audiofile/pull/124).

Beside the fix, this pull request also adds two tests that fail for the current implementation in `main`.

### DIFF
--- a/audinterface/core/utils.py
+++ b/audinterface/core/utils.py
@@ -133,21 +133,23 @@ def read_audio(
     if root is not None and not os.path.isabs(file):
         file = os.path.join(root, file)
 
+    sampling_rate = af.sampling_rate(file)
+
     if start is None or pd.isna(start):
         offset = 0
     else:
-        offset = start.total_seconds()
+        offset = to_samples(start, sampling_rate)
 
     if end is None or pd.isna(end):
         duration = None
     else:
-        duration = end.total_seconds() - offset
+        duration = to_samples(end, sampling_rate) - offset
 
     signal, sampling_rate = af.read(
         audeer.safe_path(file),
         always_2d=True,
-        offset=offset,
-        duration=duration,
+        offset=str(offset),  # str() to enforce samples as unit
+        duration=str(duration),
     )
 
     return signal, sampling_rate
@@ -162,9 +164,9 @@ def segment_to_indices(
     if pd.isna(end):
         end = pd.to_timedelta(signal.shape[-1] / sampling_rate, unit='s')
     max_i = signal.shape[-1]
-    start_i = int(round(start.total_seconds() * sampling_rate))
+    start_i = to_samples(start, sampling_rate)
     start_i = min(start_i, max_i)
-    end_i = int(round(end.total_seconds() * sampling_rate))
+    end_i = to_samples(end, sampling_rate)
     end_i = min(end_i, max_i)
     return start_i, end_i
 
@@ -385,6 +387,23 @@ def to_array(value: typing.Any) -> np.ndarray:
         elif is_scalar(value):
             value = np.array([value])
     return value
+
+
+def to_samples(
+        duration: pd.Timedelta,
+        sampling_rate: int,
+) -> int:
+    r"""Convert duration to samples.
+
+    Args:
+        duration: duration in seconds
+        sampling_rate: sampling rate in Hz
+
+    Returns:
+        duration in samples
+
+    """
+    return int(round(duration.total_seconds() * sampling_rate))
 
 
 def to_timedelta(
--- a/audinterface/core/utils.py
+++ b/audinterface/core/utils.py
@@ -133,23 +133,21 @@ def read_audio(
     if root is not None and not os.path.isabs(file):
         file = os.path.join(root, file)
 
-    sampling_rate = af.sampling_rate(file)
-
     if start is None or pd.isna(start):
         offset = 0
     else:
-        offset = to_samples(start, sampling_rate)
+        offset = start.total_seconds()
 
     if end is None or pd.isna(end):
         duration = None
     else:
-        duration = to_samples(end, sampling_rate) - offset
+        duration = end.total_seconds() - offset
 
     signal, sampling_rate = af.read(
         audeer.safe_path(file),
         always_2d=True,
-        offset=str(offset),  # str() to enforce samples as unit
-        duration=str(duration),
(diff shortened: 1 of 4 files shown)


hagenw:
As a next step we should also check what is actually the desired way to round when converting to samples:

* `int(ceil(duration * sampling_rate))`
*  `int(round(duration * sampling_rate))`
*  `int(floor(duration * sampling_rate))`

For `round()` there are also different implementations, some rounding 0.5 to 0, others to 1.

Maybe we should check what `soundfile` (and maybe devAIce)  does and if we agree with their behavior, see if we manage to adjust `audiofile` and `audinterface` to behave in the same way.

hagenw:
I decided to first update/fix `audiofile` in https://github.com/audeering/audiofile/pull/124.

There we also introduce an internal `to_samples()` function, that we could reuse here.

hagenw:
> Maybe we should check what `soundfile` (and maybe devAIce) does and if we agree with their behavior, see if we manage to adjust `audiofile` and `audinterface` to behave in the same way.

`soundfile` does require that the `offset` and `duration` (or better `start` and `end`) are given already as samples.
devAIce is discussed in https://gitlab.audeering.com/project/devaice/meta/-/issues/169

audinterface/core/utils.py
```diff
@@ -391,6 +393,23 @@ def to_array(value: typing.Any) -> np.ndarray:
...
+
+    Args:
+        duration: duration in seconds
+        sampling_rate: sampling rate in Hz
+
+    Returns:
+        duration in samples
+
+    """
+    return int(round(duration.total_seconds() * sampling_rate))
```
schruefer, line 410: In our case round() should always return an integer, so we do not necessarily need the cast.

round: "The return value is an integer if ndigits is omitted or None."

audinterface/core/utils.py
```diff
@@ -137,21 +137,23 @@ def read_audio(
...
+        duration = to_samples(end, sampling_rate) - offset
 
     signal, sampling_rate = af.read(
         audeer.safe_path(file),
         always_2d=True,
-        offset=offset,
-        duration=duration,
+        offset=str(offset),  # str() to enforce samples as unit
+        duration=str(duration),
     )
```
schruefer, line 157: The offset/duration was previously given in seconds and now is given as number of samples.
I saw that audiofile.read and respectively audmath.duration_in_seconds treats integers as seconds and strings as samples.
I wonder what the deeper sense behind it is? Without deeper knowledge it looks a bit hacky to me.

tests/test_process.py
```diff
@@ -339,8 +365,13 @@ def test_process_file(
...
         end=end,
     )
+    if isinstance(y.values, np.ndarray):
+        values = y.values[0]
+    else:
+        values = y.values
+
     np.testing.assert_almost_equal(
-        y.values, expected_output, decimal=4,
+        values, expected_output, decimal=4,
```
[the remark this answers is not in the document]
hagenw: You are totally right, there is no need to change this.
Cannot remember what motivated me to do it in the first place ;)

I removed the changes and it should look as in `main` now.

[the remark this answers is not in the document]
hagenw: Ah, ok, that was then the reason why I added my complicated code in the first place.

The reason why it worked before was that:
```python
np.testing.assert_almost_equal([1.], 1)
```
does not raise an error.
Now we compare 
```python
np.testing.assert_almost_equal(1., 1)
```
for single values.

audinterface/core/utils.py
hagenw: This line is no longer part of the updated code as we use `audmath.samples()` and do no longer require `to_samples()`.

audinterface/core/utils.py
hagenw: The idea behind the strange string option is to provide the possibility to use units like `'3 ms'` or provide samples without the need of having an additional `unit` argument.

The only problem was that we designed `audmath.duration_in_seconds()` in a way that is not compatible with how we can do it in `audiofile`:

> if `duration` is a float, integer or string without unit it is treated as seconds or if `sampling_rate` is provided as samples

The problem is in `audiofile.read()` we always have the sampling rate and hence we treat `'1000'` as representing 1000 samples instead of 1000 seconds. So yes, it's not perfect, but you also can just ignore it if you don't want to use it ;)

---

As we have now fixed `audiofile` we no longer need to change the code here in `audinterface`.

schruefer: I see, yes, all fine, i was just thinking in the first place, that this results in errors as we previously used seconds and now use samples and was quite surprised to see, that `1000` is handled differently than `"1000"`.

PR #146, opened 2023-07-28
Release 1.1.0

hagenw:
One new feature (that's why I decided for version 1.1.0) and again a few fixes for precision and order of `start` and `end` values for some rare use-cases.

image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,14 +29,25 @@ Version 1.1.0 (2023/07/28)
   in the index returned by ``audinterface.Segment``
   processing functions
 * Fixed: precision of ``audinterface.utils.to_timedelta()``
-  for ``pd.Timedelta`` objects as input
-* Fixed: precision of ``start`` and ``end`` values
-  as returned by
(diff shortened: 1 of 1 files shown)


hagenw:
@schruefer if you have trouble understanding changelog entries, please complain and we can then discuss how to improve them that a person that was not implementing the changes can understand them.

hagenw:
Good point, I extended now the two last entries.

With the first I mean indeed precision when nanoseconds are given:

image

The second case is slightly different and not affected by `to_timedelta()`. If you called `process_file()`, `process_files()`, or `process_folder()` with `start` and `end` values to process the incoming file(s), it rounded `start` and `end` in the index of the returned results to the time matching the nearest sample

image

schruefer:
Thanks for the clarification, everything sounds plausible to me now.

For the last instance, could we say "preserve requested start and end values in prcess_file()..." instead of "preserve precision of requested start and end values in prcess_file()..."?

hagenw:
Good idea, I changed accordingly.

PR #157, opened 2024-01-26
Add per execution process_func_args argument

### LINKED ISSUE #156: Suggested feature: `process_func_args` in `process_index()`etc.
```python
import audb
import audinterface
import numpy as np


def compute_stat(signal, sampling_rate, functional=np.mean):
    return functional(signal)


media = [
    'wav/03a01Fa.wav',
    'wav/03a01Nc.wav',
    'wav/16b10Wb.wav',
]
db = audb.load(
    'emodb',
    version='1.4.1',
    media=media,
)

for f in [np.mean, np.std]:
    interface = audinterface.Process(process_func=compute_stat,
                                     process_func_args={'functional': f})
    result = interface.process_index(db['emotion'].index)
    print(result)
```

If the behavior of the processing function can be controlled by arguments, it would be favorable to have an argument `process_func_args` also in `.process_index()`, `.process_folder()` etc.

The advantage might not become clear with this simple example, but consider a case where different processing functions are defined and corresponding `audinterface` objects are passed to other functions, where different indexes are processed with different `process_func_args`.

@hagenw

hagenw:
Closes #156 

This adds a `process_func_args` argument to all the processing methods, e.g. `audinterface.Process.process_index()`.
It can be used to overwrite the `process_func_args` argument used during instantiation of the interface.

Example:

```python
import audinterface
import numpy as np

def addition(signal, sampling_rate, value=1):
    return signal + value * signal

interface = audinterface.Process(
    process_func=addition,
    process_func_args={"value": 2},
)
signal = np.ones((1, 3))
sampling_rate = 8000
interface.process_signal(signal, sampling_rate, process_func_args={"value": 3})
```
returns
```
start   end                   
0 days  0 days 00:00:00.000375    [[4.0, 4.0, 4.0]]
dtype: object
```

It will not overwrite the content of the class attribute `interface.process_func_args`, which will always only store the global setting.

Example docstring:

image

### DIFF
--- a/audinterface/core/process.py
+++ b/audinterface/core/process.py
@@ -171,26 +171,8 @@ class Process:
         if win_dur is not None and hop_dur is None:
             hop_dur = utils.to_timedelta(win_dur, sampling_rate) / 2
 
-        self._process_func_special_args = {
-            'idx': False,
-            'root': False,
-            'file': False,
-        }
-        r"""Special processing function arguments.
-
-        If present in the processing function,
-        those are automatically set
-        if not included in ``process_func_args``.
-
-        """
-
-        self._process_func_args = inspect.signature(process_func).parameters
-        r"""Arguments present in the processing function."""
-
-        # Set value to ``True`` in ``self._process_func_special_args``
-        # that are not contained in ``process_func_args``,
-        # but inside ``self._process_func_args``
-        self._update_special_process_func_args(process_func_args)
+        self._process_func_signature = inspect.signature(process_func).parameters
+        r"""Arguments present in processing function."""
 
         self.channels = channels
         r"""Channel selection."""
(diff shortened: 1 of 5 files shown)


maxschmitt (approved):
Thanks for this feature and the extensive new tests.
Looks good to me.

audinterface/core/process.py
```diff
@@ -171,21 +171,9 @@ def process_func(signal, _):
...
-            'file': False,
-        }
-        for key in self._process_func_special_args:
-            if (
-                    key in signature.parameters
-                    and key not in process_func_args
-            ):
-                self._process_func_special_args[key] = True
+        self._process_func_signature = signature.parameters
+        r"""Arguments present in processing function."""
```
hagenw, line 176: @frankenjoe 

If you are curious, I think this might be the only change in `audinterface` 1.2.0, that touches attributes of the objects.

PR #163, opened 2024-03-25
Fix multiprocessing for Process, Feature, Segment

### LINKED ISSUE #165: Multiprocessing is broken at the moment
Using `multiprocessing=True` is and was broken since some time, see https://github.com/audeering/opensmile-python/issues/104 and https://github.com/audeering/audinterface/pull/163.

hagenw:
Closes #165 

Ensure, that the `multiprocessing` argument of `audinterface.Process`, `audinterface.Feature`, `audinterface.Segment` does not result in an pickle related error, if set to `True`. Multiprocessing relies on [pickling all involved objects](https://laszukdawid.com/blog/2017/12/13/multiprocessing-in-python-all-about-pickling/) to communicate between processes.
The error was originally reported in https://github.com/audeering/opensmile-python/issues/104.

I added tests for `audinterface.Process`, `audinterface.Feature`, `audinterface.Segment`, to detect the error.
Most tests included already the `multiprocessing` argument, but it was always set to `False`.

The reason for the error consisted of two parts:

1. In https://github.com/audeering/audinterface/pull/157 we introduced
```python
signature = inspect.signature(process_func)
Process._process_func_signature = signature.parameters
```
The problem is that `Process._process_func_signature` cannot be pickled.
To fix this, I changed the code to 
```python
signature = inspect.signature(process_func)
Process._process_func_signature = dict(signature.parameters)
```
2. When `process_func=None` is selected, we provide default process functions. Before they were defined on the fly inside the object, which also results in objects that cannot be pickled. This is solved here, by defining the functions outside of the actual objects.

/cc @frankenjoe

### DIFF
--- a/audinterface/core/feature.py
+++ b/audinterface/core/feature.py
@@ -16,6 +16,30 @@ from audinterface.core.typing import Timestamps
 import audinterface.core.utils as utils
 
 
+def zeros(signal, sampling_rate, num_channels, num_features) -> np.ndarray:
+    r"""Default feature process function.
+
+    This function is used,
+    when ``Feature`` is instantiated
+    with ``process_func=None``.
+    It returns zeros for all channels and features.
+
+    Args:
+        signal: signal
+        sampling_rate: sampling rate in Hz
+        num_channels: number of audio channels
+        num_features: number of features
+
+    Returns:
+        zeros with size ``(num_channels, num_features)``
+
+    """
+    return np.zeros(
+        (num_channels, num_features),
+        dtype=object,
+    )
+
+
 class Feature:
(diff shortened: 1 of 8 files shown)


hagenw:
Note, this test does not fail for `audinterface` v1.1.0

hagenw:
This error happens when using `process_func=None`. But not if using another processing function defined inside the test script, e.g. `identity()`.

hagenw:
The error can be solved by not using `process_func=None` until pull request https://github.com/audeering/audinterface/pull/157, afterwards that does not help.

maxschmitt (approved):
Looks good to me, thanks for the fix.

PR #170, opened 2024-03-27
Release 1.2.1

hagenw:
image


maxschmitt (approved):
Sounds clear to me, though I am not a big fan of changing default parameters.

hagenw:
> though I am not a big fan of changing default parameters.

Yes, this is indeed not a good idea. In this case it was announced ~1 year ago with a deprecation warning, that the default will change with version 1.2.0 of `audinterface`, but I forgot to do it. Hence it is included in this bug fix release.

PR #174, opened 2024-05-16
Test fix in audb

hagenw:
Depend on https://github.com/audeering/audb/pull/411 to show that the fix proposed in `audb` does solve the [error we are seeing here for Python 3.8](https://github.com/audeering/audinterface/actions/runs/9094046905/job/25009555875?pr=172).


hagenw:
No longer needed.

PR #186, opened 2024-12-16
Release 1.2.3

hagenw:
image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,13 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
+Version 1.2.3 (2024/12/16)
+--------------------------
+
+* Added: support for Python 3.12
+* Changed: update progress bar at least every second
+
+
 Version 1.2.2 (2024/06/28)
 --------------------------
 


ChristianGeng (commented):
I think the `sibyll` MR was merged later than the 1.2.2 release:

https://github.com/audeering/audinterface/pull/178

hagenw:
Yes, but I did not mention `sybil` as it should not effect the user.

The user is slightly effected as the documentation now looks slightly different. I could add the following:
```
* Changed: use doctest syntax for showing code examples in the documentation
```

Would this be better than stating nothing?

ChristianGeng:
> Yes, but I did not mention `sybil` as it should not effect the user.
> 
> The user is slightly effected as the documentation now looks slightly different. I could add the following:
> 
> ```
> * Changed: use doctest syntax for showing code examples in the documentation
> ```
> 
> Would this be better than stating nothing?

Either: I took the wrong perspective. You are right in saying that it does not affect the user how the docstrings are rendered.  So either of them is fine with me. I will proceed with approval.

ChristianGeng (approved):
I only had a single minor issue about rendering documentation.
This is settled and I am approving this.

PR #189, opened 2025-05-22
Add SegmentWithFeature interface

### LINKED ISSUE #188: Segment with features
In some situations, it might make sense to apply segmentation and extract features at the same time.

For example, if you want to apply ASR and get both the time stamps and the contained words, you wouldn't first apply word-level segmentation and then ASR, but many models provide both time stamps and transcripts at the same time.
Another example would be an external function that internally applies voice activity detection and then speech classification (e.g. devAIce). 

Since I needed this type of interface for some tasks, I added it as a class `SegmentWithFeatures` (a sort of combination of `audinterface.Segment` and `audinterface.Feature`) in my own code, and so far it seems to work as intended.

Would this be worth adding to `audinterface`, or do you see reasons against it?

@hagenw

audeerington:
Closes #188 

This adds `audinterface.SegmentWithFeature`, for process functions that apply segmentation and feature extraction at the same time.

image

Usage documentation:

image

## Summary by Sourcery

Add a new SegmentWithFeature class to audinterface that unifies audio segmentation and feature extraction workflows, complete with default behavior, multi-channel support, documentation updates, and extensive tests.

New Features:
- Introduce SegmentWithFeature interface for combined segmentation and feature extraction
- Provide default empty_series function when no process_func is supplied

Enhancements:
- Support processing via process_file, process_folder, process_index, process_table, process_signal, and process_signal_from_index methods
- Implement internal reshaping and conversion utilities for multi-channel and multi-feature outputs

Build:
- Add faster-whisper to test dependencies

Documentation:
- Update usage documentation to include the new SegmentWithFeature interface

Tests:
- Add comprehensive test_segment_with_feature.py covering all methods and edge cases

## Summary by Sourcery

Add a new SegmentWithFeature interface to audinterface that unifies segmentation and feature extraction into a single, flexible API. Include multi-channel support, default behaviors, extended processing methods (file, folder, index, signal, table), caching, and label propagation, along with updated documentation and extensive tests.

New Features:
- Introduce SegmentWithFeature interface for combined segmentation and feature extraction workflows
- Provide a default empty_series process function when none is specified

Enhancements:
- Support multi-channel inputs and channel selection in SegmentWithFeature
- Add methods for processing individual files, lists of files, folders, signal arrays, segmented indices, and tables with optional caching and label propagation
- Implement internal reshaping and conversion utilities for consistent feature output

Build:
- Add faster-whisper to test dependencies

Documentation:
- Update usage and API documentation to cover the new SegmentWithFeature interface

Tests:
- Add comprehensive tests for SegmentWithFeature covering all methods, edge cases, and data types

### DIFF
--- a/audinterface/core/segment_with_feature.py
+++ b/audinterface/core/segment_with_feature.py
@@ -120,13 +120,11 @@ class SegmentWithFeature:
         ...     ends = pd.to_timedelta(
         ...         np.arange(win_size, size + (1 / sampling_rate), hop_size), unit="s"
         ...     )
-        ...     # Get windows of shape (samples, frames)
-        ...     frames = utils.sliding_window(signal, sampling_rate, win_size, hop_size)[0]
-        ...     frames = frames.transpose(1, 0)
-        ...     means = frames.mean(axis=1)
-        ...     stds = frames.std(axis=1)
+        ...     # Get windows of shape (channels, samples, frames)
+        ...     frames = utils.sliding_window(signal, sampling_rate, win_size, hop_size)
+        ...     means = frames.mean(axis=(0, 1))
+        ...     stds = frames.std(axis=(0, 1))
         ...     index = pd.MultiIndex.from_tuples(zip(starts, ends), names=["start", "end"])
-        ...     # Pass list of arrays with shape (channels, features) to create series
         ...     features = list(np.stack((means, stds), axis=-1))
         ...     return pd.Series(data=features, index=index)
         >>> interface = SegmentWithFeature(
--- a/audinterface/core/segment_with_feature.py
+++ b/audinterface/core/segment_with_feature.py
@@ -670,7 +670,7 @@ class SegmentWithFeature:
             self._check_return_format(series)
             if series.empty:
                 continue
-            data = [self._reshape_numpy_1d(values) for values in series]
+            data = self._reshape_numpy(series)
             data = np.stack(data).T
             index_data = series.index.to_numpy()
             files.extend([file] * len(series))
@@ -725,7 +725,7 @@ class SegmentWithFeature:
             self._check_return_format(series)
             if series.empty:
                 continue
-            data = [self._reshape_numpy_1d(values) for values in series]
+            data = self._reshape_numpy(series)
             data = np.stack(data).T
             index_data = series.index.to_numpy()
             files.extend([file] * len(series))
@@ -753,7 +753,7 @@ class SegmentWithFeature:
             self._check_return_format(series)
             if series.empty:
                 continue
-            data = [self._reshape_numpy_1d(values) for values in series]
+            data = self._reshape_numpy(series)
             data = np.stack(data).T
             index_data = series.index.to_numpy()
             starts.extend([t[0] + start for t in index_data])
@@ -769,6 +769,23 @@ class SegmentWithFeature:
             data=features,
             columns=self.feature_names,
         )
+
+    def _reshape_numpy(
+        self,
+        features: pd.Series,
+    ):
+        r"""Reshape values in series to numpy array of shape [n_features]."""
+        # Cover case that process function returned a series
+        # with numpy arrays as elements
+        if pd.api.types.is_object_dtype(features):
+            data = [self._reshape_numpy_1d(values) for values in features]
+        else:
+            # Case that the process function returned a series
+            # with elements that are scalar values.
+            # Use to_numpy() directly to preserve dtype of series
+            data = features.to_numpy()
+            data = np.reshape(data, (len(features), self.num_features))
+        return data
 
     def _reshape_numpy_1d(
         self,
--- a/audinterface/core/segment_with_feature.py
+++ b/audinterface/core/segment_with_feature.py
@@ -259,22 +259,14 @@ class SegmentWithFeature:
         """
         if start is None or pd.isna(start):
             start = pd.to_timedelta(0)
-        series = self.process.process_file(
+        y = self.process.process_file(
             file,
             start=start,
             end=end,
             root=root,
             process_func_args=process_func_args,
-        ).values[0]
-        self._check_return_format(series)
-        df = self._series_to_frame(series)
-        index = audformat.segmented_index(
-            files=[file] * len(df),
-            starts=df.index.get_level_values("start") + start,
-            ends=df.index.get_level_values("end") + start,
         )
-        df.index = index
-        return df
+        return self._construct_frame_segmented_index(y)
 
     def process_files(
         self,
@@ -326,28 +318,7 @@ class SegmentWithFeature:
             root=root,
             process_func_args=process_func_args,
         )
-        files = []
-        starts = []
-        ends = []
-        features = {col: [] for col in self.feature_names}
-        for (file, start, _), series in y.items():
-            self._check_return_format(series)
-            df = self._series_to_frame(series)
-            files.extend([file] * len(df))
-            starts.extend(df.index.get_level_values("start") + start)
-            ends.extend(df.index.get_level_values("end") + start)
-            for col in self.feature_names:
-                features[col].extend(df[col])
-        if len(files) == 0:
-            # Pass no data to ensure consistent dtype for columns
-            return pd.DataFrame(
-                index=audformat.segmented_index(), columns=self.feature_names
-            )
-        return pd.DataFrame(
-            index=audformat.segmented_index(files, starts, ends),
-            data=features,
-            columns=self.feature_names,
-        )
+        return self._construct_frame_segmented_index(y)
 
     def process_folder(
         self,
@@ -458,29 +429,7 @@ class SegmentWithFeature:
             cache_root=cache_root,
             process_func_args=process_func_args,
         )
-
-        files = []
-        starts = []
-        ends = []
-        features = {col: [] for col in self.feature_names}
-        for (file, start, _), series in y.items():
-            self._check_return_format(series)
-            df = self._series_to_frame(series)
-            files.extend([file] * len(df))
-            starts.extend(df.index.get_level_values("start") + start)
-            ends.extend(df.index.get_level_values("end") + start)
-            for col in self.feature_names:
-                features[col].extend(df[col])
-        if not files:
-            # Pass no data to ensure consistent dtype for columns
-            return pd.DataFrame(
-                index=audformat.segmented_index(), columns=self.feature_names
-            )
-        return pd.DataFrame(
-            index=audformat.segmented_index(files, starts, ends),
-            data=features,
-            columns=self.feature_names,
-        )
+        return self._construct_frame_segmented_index(y)
 
     def process_signal(
         self,
@@ -527,39 +476,18 @@ class SegmentWithFeature:
         .. _audformat: https://audeering.github.io/audformat/data-format.html
 
         """
-        series = self.process.process_signal(
+        y = self.process.process_signal(
             signal,
             sampling_rate,
             file=file,
             start=start,
             end=end,
             process_func_args=process_func_args,
-        ).values[0]
-        self._check_return_format(series)
-        index = series.index
-        df = self._series_to_frame(series)
-        if start is not None:
-            start = utils.to_timedelta(start)
-            # Here we change directly the levels,
-            # so we need to use
-            # `index.levels[0]`
-            # instead of
-            # `index.get_level_values('start')`
-            index = index.set_levels(
-                [
-                    index.levels[0] + start,
-                    index.levels[1] + start,
-                ],
-                level=[0, 1],
-            )
-        if file is not None:
-            index = audformat.segmented_index(
-                files=[file] * len(index),
-                starts=index.get_level_values("start"),
-                ends=index.get_level_values("end"),
-            )
-        df.index = index
-        return df
+        )
+        if file is None:
+            return self._construct_frame_signal_index(y)
+        else:
+            return self._construct_frame_segmented_index(y)
 
     def process_signal_from_index(
         self,
@@ -634,7 +562,6 @@ class SegmentWithFeature:
             task_description=f"Process {len(index)} segments",
             maximum_refresh_time=1,
         )
-
         files = []
         starts = []
         ends = []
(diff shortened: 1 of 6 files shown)


audeerington:
@hagenw I added a draft for #189 here, but the tests fail for a usage example that I didn't change. The results for the computed `f0` are different than what is expected. Maybe because of a different `librosa` version?

audeerington:
> audinterface is already quite complex and hard to maintain, so one solution would be to not add process_table().

True, I guess we should go with a solution that doesn't add many more complications. I would slightly prefer the option of returning two dataframes then, which seems like the cleanest option in terms of handling the overlapping columns and different column levels, since I do see some use cases for having this function.

> Do you know how audinterface.Segment handles this?

As I understand, `audinterface.Segment` only returns an index, so one doesn't need to cover different column levels. But there isn't really a way to differentiate that a certain segment originates from a certain channel, I think.

hagenw:
> I guess we should go with a solution that doesn't add many more complications. I would slightly prefer the option of returning two dataframes then

> `audinterface.Segment` also doesn't have a `process_func_is_mono` argument, and since the processing functions have to return a single index and no features, the problem never comes up.

One solution would be to also only support mono here. Instead of returning multiple dataframes, the user would then create different interfaces for different input channels. The only downside I see would be that it has to read the files then several times.

> handling the overlapping columns

In that case we could also raise an error and ask the user to first rename the affected columns?

hagenw:
> I would slightly prefer the option of returning two dataframes

The downside is that this introduces another behavior. At the moment `Segment` does only support mono, and `Feature` does return a multiindex as column names. I would expect that `SegmentWithFeature` follows one of those two approaches.

hagenw:
Great, then I would propose you continue with the suggested changes.

tests/test_segment_with_feature.py
```diff
@@ -0,0 +1,1108 @@
...
+@pytest.mark.parametrize(
+    "signal, sampling_rate, segment_with_feature, expected",
+    [
+        (
+            ONES_1D,
+            SAMPLING_RATE,
+            audinterface.SegmentWithFeature(
+                feature_names="feature",
+                process_func=None,
+            ),
```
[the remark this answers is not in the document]
audeerington: I added some test cases to `test_call` that cover the case that the process function output has the wrong format.

tests/test_segment_with_feature.py
```diff
@@ -0,0 +1,1108 @@
...
+def test_file(tmpdir, num_channels, num_features):
+    feature_names = [f"f{i}" for i in range(num_features)]
+    segmentwithfeature = audinterface.SegmentWithFeature(
+        feature_names=feature_names,
+        process_func=lambda x, sr: pd.Series(
+            data=[np.ones(num_features)] * len(INDEX), index=INDEX
+        ),
+        sampling_rate=None,
+        channels=range(num_channels),
+        resample=False,
```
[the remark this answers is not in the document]
audeerington: I added tests similar to the suggestion to `test_signal` that cover minimum and maximum signal duration.

tests/test_segment_with_feature.py
```diff
@@ -0,0 +1,1108 @@
...
+                process_func=segment_non_zeros_with_mean_mono,
+                channels=range(2),
+            ),
+        ),
+    ],
+)
+def test_call_datatype(signal, extractor):
+    features = extractor(signal, SAMPLING_RATE)
+    for feature in features:
+        assert signal.dtype == feature.dtype
```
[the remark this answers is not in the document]
audeerington: Removed loops and conditionals from the tests.

audeerington:
@hagenw Sorry if you got many update notifications from this, I wasn't expecting the AI generated review and I added you as reviewer too soon. Anyway, I think the generated review comments are all addressed now and it is ready.

audinterface/core/segment_with_feature.py
```diff
@@ -0,0 +1,825 @@
...
+        >>> interface = SegmentWithFeature(
+        ...     feature_names=["mean", "std"], process_func=segment_with_mean_std
+        ... )
+        >>> signal = np.array([1.0, 2.0, 3.0, 4.0, 5.0, 6.0])
+        >>> interface(signal, sampling_rate=2)
+        start            end
+        0 days 00:00:00  0 days 00:00:01    [1.5, 0.5]
+        0 days 00:00:01  0 days 00:00:02    [3.5, 0.5]
+        0 days 00:00:02  0 days 00:00:03    [5.5, 0.5]
+        dtype: object
```
hagenw, line 139: Using a process function with a fixed win and hop size is also supported by `audinterface.Feature`. Could you maybe do a benchmark in terms of execution time on a dataset (e.g. emodb) comparing your process function for `audinterface.Feature` and `audinterface.SegmentWithFeature` to double check that the new implementation is not too slow. Note, for `audinterface.Feature` the process function just needs to return the features.

docs/usage.rst
```diff
@@ -441,6 +441,83 @@ wav/16b10Wb.wav 0 days 00:00:00.300000 0 days 00:00:01             4
...
+            for word in segment.words:
+                index.append(
+                    (
+                        pd.to_timedelta(word.start, unit="s"),
+                        pd.to_timedelta(word.end, unit="s")
+                    )
+                )
+                words.append(word.word.strip())
+        index = pd.MultiIndex.from_tuples(index, names=["start", "end"])
+        return pd.Series(data=words, index=index)
```
hagenw, line 482: I tried to change this line to

```python
        return pd.Series(data=words, index=index, dtype="string")
```

but the returned dataframe of `interface.process_file(...)` has still `dtype=object` for the `word` column, which seems not so nice to me. How are dtypes handled in general in the code?

audeerington: Good point...

I tried to stay as close to `audinterface.Feature` as possible. So currently each element of the series is converted to a numpy array first (if it isn't a numpy array already), and then reshaped to ensure it has the expected number of features. And when the numpy data is used to create the resulting dataframe, `dtype=object` is used:
```python
>>> feats = np.array(["a", "b", "c"])
>>> feats.dtype
dtype('<U1')
>>> frame = pd.DataFrame({"feat": feats})
>>> frame["feat"].dtype
dtype('O')
```
If you have a processing function for `audinterface.Feature` that returns a string, the result will also have `dtype=object`.

An alternative solution could be that the processing function has to return a dataframe with correctly named columns and with dtype already set.

hagenw: > If you have a processing function for `audinterface.Feature` that returns a string, the result will also have `dtype=object`.

I haven't checked this, but this is a valid point. Until we haven't changed/fixed this in `audinterface.Feature` it makes sense that `audinterface.SegmentWithFeature` behaves the same way.

audinterface/core/segment_with_feature.py
audeerington: Ok, I ran a benchmark on emodb to compare the mean/std segments process function, with a window size of 0.5 and a hop size of 0.25. The `audinterface.SegmentWithFeature` implementation takes about twice as long compared to `audinterface.Feature`.

<details>
<summary>Benchmarking script</summary>

```python
from time import perf_counter

import numpy as np
import pandas as pd

import audb

import audinterface


db = audb.load(
    "emodb", version="1.4.1", format="wav", sampling_rate=16000, mixdown=True
)

df = db["files"].get()

WIN_SIZE = 0.5
HOP_SIZE = 0.25


def segment_with_mean_std(
    signal, sampling_rate, *, win_size=WIN_SIZE, hop_size=HOP_SIZE
):
    size = signal.shape[1] / sampling_rate
    starts = pd.to_timedelta(
        np.arange(0, size - win_size + (1 / sampling_rate), hop_size),
        unit="s",
    )
    ends = pd.to_timedelta(
        np.arange(win_size, size + (1 / sampling_rate), hop_size), unit="s"
    )
    # Get windows of shape (channels, samples, frames)
        signal, sampling_rate, win_size, hop_size
    )
    means = frames.mean(axis=(0, 1))
    stds = frames.std(axis=(0, 1))
    index = pd.MultiIndex.from_tuples(zip(starts, ends), names=["start", "end"])
    features = list(np.stack((means, stds), axis=-1))
    return pd.Series(data=features, index=index)


def mean_std(signal, sampling_rate):
    return [signal.mean(), signal.std()]


swf_interface = audinterface.SegmentWithFeature(
    feature_names=["mean", "std"], process_func=segment_with_mean_std
)


f_interface = audinterface.Feature(
    feature_names=["mean", "std"],
    process_func=mean_std,
    win_dur=WIN_SIZE,
    hop_dur=HOP_SIZE,
)

for name, interface in [
    ("audinterface.Feature", f_interface),
    ("audinterface.SegmentWithFeature", swf_interface),
]:
    n = 10
    durations = []
    for i in range(n):
        time_start = perf_counter()
        interface.process_index(df.index)
        time_end = perf_counter()
        duration = time_end - time_start
        durations.append(duration)
    average_duration = sum(durations) / len(durations)
    print(f"{name}: Average elapsed time over {n} repetitions: {duration:.3f} seconds")

```
</details>

```
audinterface.Feature: Average elapsed time over 10 repetitions: 0.281 seconds                       
audinterface.SegmentWithFeature: Average elapsed time over 10 repetitions: 0.590 seconds
```

audeerington: > (it looks like you iterate over a series that contain other series, is that correct?).

Yes. The `process` attribute returns a series with the results of each process function call, which is always a series. So we have this nested series and need to turn it into a dataframe.

This is the same for `_construct_frame_signal_index()` and for `_construct_frame_segmented_index()` - depending on whether `process` creates a signal index or a segmented index, the respective helper function is used.

audinterface/core/segment_with_feature.py
```diff
@@ -0,0 +1,825 @@
...
+        starts = []
+        ends = []
+        features = {col: [] for col in self.feature_names}
+        for (start, _), series in y.items():
+            self._check_return_format(series)
+            df = self._series_to_frame(series)
+            starts.extend(df.index.get_level_values("start") + start)
+            ends.extend(df.index.get_level_values("end") + start)
+            for col in self.feature_names:
+                features[col].extend(df[col])
```
hagenw, line 751: Maybe you can try to avoid creating additional dataframes and calling `get_level_values()` by something similar to this:

```suggestion
        for (start, _), series in y.items():
            self._check_return_format(series)
            if series.empty:
                data = np.empty((1, len(self.feature_names)))
            else:
                data = [self._reshape_numpy_1d(values) for values in series]
                data = np.stack(data).T
            starts.extend(series.index.levels[0] + start)
            ends.extend(series.index.levels[1] + start)
            for col, val in zip(self.feature_names, data):
                features[col].extend(val)
```

I have not tested if this is really faster, but I would assume it should be.

hagenw: The following code should work instead:

```suggestion
        for (start, _), series in y.items():
            self._check_return_format(series)
            if series.empty:
                data = np.empty((1, len(self.feature_names)))
            else:
                data = [self._reshape_numpy_1d(values) for values in series]
                data = np.stack(data).T
            index_data = series.index.to_numpy()
            if len(index_data) > 0:
                starts.extend([t[0] + start for t in index_data])
                ends.extend([t[1] + start for t in index_data])
            for col, val in zip(self.feature_names, data):
                features[col].extend(val)
```

I updated `_construct_frame_segmented_index()` in the same way, and just got one failing test:

```
E       AssertionError: Attributes of DataFrame.iloc[:, 0] (column name="mean") are different
E       
E       Attribute "dtype" are different
E       [left]:  float32
E       [right]: float64
```

I then first rerun your benchmark with my changes:

```
audinterface.Feature: Average elapsed time over 100 repetitions: 0.385 seconds                      
audinterface.SegmentWithFeature: Average elapsed time over 100 repetitions: 0.744 seconds
```

and repeated it with my changes applied:

```
audinterface.Feature: Average elapsed time over 100 repetitions: 0.378 seconds                      
audinterface.SegmentWithFeature: Average elapsed time over 100 repetitions: 0.692 seconds
```

Unfortunately, this does not really help much.

hagenw: Maybe, we just stay with your proposed implementation?

audeerington: I integrated the suggestion, and adjusted the tests. The failing test revealed that the returned dtype wasn't kept when constructing the dataframe before, and that the suggestion was more correct than the previous implementation :)

I adjusted and added some tests to check for the correct handling of float dtypes, and added the helper `_reshape_numpy()` to cover the case that the returned series contains scalar values, so we still preserve the dtype in that case.

I also reran the benchmark script again, and there is a similar improvement in execution time to before.

tests/test_segment_with_feature.py
```diff
@@ -0,0 +1,1279 @@
...
+    non_zero_index = np.where(signal_1d != 0)[0]
+    if len(non_zero_index) == 0:
+        return pd.Series(index=audinterface.utils.signal_index([], []))
+    # Find points where non-zero segments stop
+    # by selecting the points where there is a gap (np.diff != 1)
+    # in the non-zero indices
+    split_indices = np.where(np.diff(non_zero_index) != 1)[0]
+    frames = np.split(signal_1d[non_zero_index], split_indices + 1)
+    frame_lengths = [len(frame) for frame in frames]
+
```
hagenw, line 48:
```suggestion
```

tests/test_segment_with_feature.py
```diff
@@ -0,0 +1,1279 @@
...
+                    data=[1.0],
+                    index=pd.MultiIndex.from_tuples(
+                        [("a", "b")], names=["start", "end"]
+                    ),
+                ),
+            ),
+            None,
+            marks=pytest.mark.xfail(raises=ValueError),
+        ),
+        # incorrectly returning dataframe instead of series
```
hagenw, line 121: Could you maybe also add a test what happens if you return only the multiindex instead of a series.

tests/test_segment_with_feature.py
```diff
@@ -0,0 +1,1279 @@
...
+            ),
+            pd.DataFrame(
+                index=audformat.segmented_index(
+                    files=["f1.wav", "f1.wav", "f2.wav"],
+                    starts=[
+                        0,
+                        2,
+                        1,
+                    ],
+                    ends=[1, 3, 2],
```
hagenw, line 577:
```suggestion
                    starts=[0, 2, 1],
                    ends=[1, 3, 2],
```

hagenw (commented):
Thanks, I have not checked every test in detail. This is anyway not so easy in `audinterface`, but I also have no easy solution how to make it easier. Basically, you never know if you missed to test for something ;)

PR #191, opened 2025-05-22
TST: require librosa>=0.11.0 for F0 extraction

hagenw:
Updates the usage example, which uses `librosa` for F0 extraction to use `librosa>=0.11.0` and adjust the output values accordingly.

image

## Summary by Sourcery

Require librosa>=0.11.0 for F0 extraction and update usage example accordingly

Documentation:
- Adjust F0 extraction usage example to match updated librosa output values

Tests:
- Bump librosa dependency to >=0.11.0 in tests requirements


audeerington (approved):
Thanks, looks good to me!

PR #204, opened 2026-07-23
Use strict index hash for caching

### LINKED ISSUE #203: Use strict hash for caching
The file name that is used for caching in processes is set to `hash = audformat.utils.hash(index)` . Unless `strict=True` is used in `audformat.utils.hash()`, the order of the index does not change the hash.

This can mean that if one has already processed an index with the same elements, but then processes the same index with a different order, one will get an unexpected order of index in the results.
Also, if one uses e.g. `process_index(index, cache_root="cache", preserve_index=True)` in this situation, the order of the returned features will be the same as the input index, but the features will be in the wrong order.

I would suggest to change to use `hash = audformat.utils.hash(index, strict=True)` to prevent this issue. However, this would mean that existing caches could not be reused when switching to a new `audinterface` version - would that be ok @hagenw ?

audeerington:
Closes #203 

Switch to using `audformat.utils.hash(index, strict=True)` for storing process results in a cache file.

This also adds a test that checks whether the returned index is as expected when the same index but with different order has already been cached.

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -740,6 +740,9 @@ def test_process_index(tmpdir, num_workers, multiprocessing, preserve_index):
 )
 @pytest.mark.parametrize("preserve_index", [False, True])
 def test_process_index_order(tmpdir, index, durations, preserve_index):
+    # Ensure the returned index is as expected
+    # when the same index but different order have already been cached
+    # https://github.com/audeering/audinterface/issues/203
     cache_root = os.path.join(tmpdir, "cache")
     process = audinterface.Process(
         process_func=None,
@@ -774,10 +777,10 @@ def test_process_index_order(tmpdir, index, durations, preserve_index):
     y_reverse = process.process_index(
         reverse_index, root=root, cache_root=cache_root, preserve_index=preserve_index
     )
+    # Make sure the index is as expected
     if preserve_index or is_segmented_index:
         expected_index = reverse_index
     else:
-        # Make sure the index is as expected
         expected_index = audformat.segmented_index(
             files=files[::-1],
             starts=[0] * len(reverse_index),
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -49,6 +49,20 @@ def signal_modification(signal, sampling_rate, subtract=False):
     return signal
 
 
+def index_durations(index: pd.Index, root: str = None) -> pd.Series:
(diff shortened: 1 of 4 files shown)


tests/test_process.py
```diff
@@ -710,6 +724,82 @@ def test_process_index(tmpdir, num_workers, multiprocessing, preserve_index):
...
+    # Make sure the index is as expected
+    if preserve_index or is_segmented_index:
+        expected_index = reverse_index
+    else:
+        expected_index = audformat.segmented_index(
+            files=files[::-1],
+            starts=[0] * len(reverse_index),
+            ends=reverse_durations.values,
+        )
+    pd.testing.assert_index_equal(y_reverse.index, expected_index)
```
hagenw, line 790: Maybe we should check here if the original index still results in the original durations, because the current test would also succeed if `process_index()` simply overwrites the existing cache instead of reading from it.

hagenw:
Unfortunately, `audinterface.Features` does not simply inherit everything from `audinterface.Process`, but has it's own code for certain things as the hashing at https://github.com/audeering/audinterface/blob/8264e46c645119388a110043dd7c5a6b03004ecf/audinterface/core/feature.py#L548

There you have to change it as well and add an extra test for `audinterface.Feature.process_index()`.

tests/test_process.py
audeerington: Hm, good point. Would it be fine to check if there exist two files in the cache directory?
Because I'm not sure how best to test this scenario. If `process_index()` overwrites the existing cache, then it would probably also overwrite the cache a second time when checking if the original index still results in the same durations.

hagenw: Yes, checking if two files exists is sufficient.

audeerington: Ok, I added this check.

audeerington:
> Unfortunately, `audinterface.Features` does not simply inherit everything from `audinterface.Process`, but has it's own code for certain things as the hashing at
> 
> https://github.com/audeering/audinterface/blob/8264e46c645119388a110043dd7c5a6b03004ecf/audinterface/core/feature.py#L548
> 
> There you have to change it as well and add an extra test for `audinterface.Feature.process_index()`.

Thanks, that would have been bad to miss! I now added a test and change for `audinterface.Feature` as well.

hagenw (approved):

PR #206, opened 2026-08-05
CI: update checkout, setup-python, setup-uv, codecov

ChristianGeng:
## Summary

Two related CI bugs, both caused by stale GitHub Action version pins:

1. **Dead uv caching**: `astral-sh/setup-uv`'s default `cache-dependency-glob`
   keys on `uv.lock`/`requirements*.txt`, neither of which exists here (no
   committed lockfile, by design), so caching never actually worked. Bumped
   `astral-sh/setup-uv` `v5` -> `v9.0.0`: `v6.0.0` added `pyproject.toml` to
   the default glob, which is committed and changes exactly when a
   dependency does, so caching now works with no lockfile needed.

2. **Node.js 20 deprecation**: `actions/checkout` and `actions/setup-python`
   bumped `v4`/`v5` -> `v7`, clearing the "Node.js 20 is deprecated" warning.
   `codecov/codecov-action` bumped `v4` -> `v7`; its `v5` rewrite dropped the
   singular `file:` input in favor of `files:`, renamed accordingly.
   `actions/cache` (used for the emodb test-data cache in doc/test
   workflows) bumped `v4` -> `v6` for the same reason.

`prune-cache` left at its new default (off): audinterface's dependency tree
(audeer, audformat, audiofile, audmath, audresample) has no large
pre-built binary wheels like torch, so pruning would save ~0 disk space
while costing avoidable re-downloads.

Part of the same CI cleanup as audeering/audeer#206,
audeering/opensmile-python#132, audeering/audb#591,
audeering/audformat#539, audeering/audbackend#307, and
audeering/audresample#83.

## Test plan

- [x] All workflow YAML files reviewed and diff-verified against the
      established pattern from sibling repos
- [ ] CI passes on this PR

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -16,21 +16,21 @@ jobs:
         python-version: [ '3.10' ]
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v7
 
     - name: Cache emodb
-      uses: actions/cache@v4
+      uses: actions/cache@v6
       with:
         path: ~/audb
         key: emodb-1.3.0
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v7
       with:
         python-version: ${{ matrix.python-version }}
 
     - name: Install uv
-      uses: astral-sh/setup-uv@v5
+      uses: astral-sh/setup-uv@v9.0.0
 
     - name: Ubuntu - install libsndfile1 and portaudio
       run: |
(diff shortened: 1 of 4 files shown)


ChristianGeng:
Re: the version-pin suggestion — pinning `astral-sh/setup-uv` to the exact `v9.0.0` tag rather than floating `v9` is intentional here, not an oversight. This bump exists specifically to land the caching fix that landed in `v6.0.0` (adding `pyproject.toml` to the default `cache-dependency-glob`), so pinning the exact version we verified against avoids silently picking up a future patch release we haven't tested. Same choice made across the sibling repos in this rollout (audeer#206, opensmile-python#132, audb#591, audformat#539, audbackend#307, audresample#83, auglib#60) — `checkout`/`setup-python`/`codecov-action` float on major tags by convention, `setup-uv` is pinned exact because this PR's whole point is a specific version's behavior change. Leaving as-is.

hagenw (approved):

